PR #3, opened 2018-01-21
New Makefile + barry_md.py refactor.

atomgalaxy:
Squashed previous two commits and therefore opened a new pull request.

Everything still the same as it ever was, except:
- now barry_md is importable
- has a --style parameter
- make, make deducing-this and make clean work as expected
- barry_md.py outputs \n on both windows and linux to avoid excessive
  diffs on the html (although we have to take this one hit to the diff)
- --help works now
- output file is not touched if parsing fails, thus not breaking make
  (unlike before, since output redirection had to be used)


atomgalaxy:
So... you gonna merge this? :)

BRevzin:
Hell yea! 👍

PR #6, opened 2018-09-26
Some more small fixes, also added a possible design for constraining …

atomgalaxy:
…to base.

### DIFF
--- a/0847_deducing_this/deducing-this.md
+++ b/0847_deducing_this/deducing-this.md
@@ -26,7 +26,7 @@ Abstract: We propose a new mechanism for specifying or deducing the value catego
 
 >  <table style="font-size=8px"><tr><th>SF</th><th>F</th><th>N</th><th>A</th><th>SA</th></tr><tr><td>0</td><td>2</td><td>9</td><td>14</td><td>12</td></table>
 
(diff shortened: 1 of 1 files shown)


0847_deducing_this/deducing-this.md
```diff
@@ -859,9 +859,15 @@ We could also introduce a new "magic" cast that just gives us a pointer to the t
 This problem extends further than just to `Self`, though. It is common to only want to deduce the ref-qualifier in all sorts of contexts. Any "make it easy to get the base class pointer"-style feature suffers extra instantiations when we only really want the instantiations for the base class. A complementary feature could be proposed that constrains *deduction* (as opposed to removing candidates once they are deduced, as with `requires`, with the following straw-man syntax:
 
     ::cpp
-    struct B { };
+    struct B {
+        template <typename Self : B>
+        auto front(this Self&& self) {
```
BRevzin: Do we want to use the old or new syntax in this section? (Or does it matter?)

atomgalaxy: "old" and "new" is going to be ambiguous pretty soon so let's call it param syntax and tail syntax? Also I think param syntax is fine here.

PR #7, opened 2018-10-03
First set of hopefully non-problematic fixes.

atomgalaxy:


0847_deducing_this/deducing-this.bs
```diff
@@ -273,13 +273,45 @@ void ex(X& x, D const& d) {
 
 Member functions with an explicit object parameter cannot be `static` or have *cv*- or *ref*-qualifiers. 
 
-A call to a member function will interpret the object parameter as the first argument to such a member function, the first argument in the parenthesized expression list as the second parameter, and so forth. 
+A call to a member function will interpret the object argument as the first (`this`-annotated) parameter to such a member function, the first argument in the parenthesized expression list as the second parameter, and so forth. 
 
 Following normal deduction rules, the template parameter corresponding to the explicit object parameter can deduce to a type derived from the class the member function is declared in (as in the example above for `d.bar()`).
 
+
+### Explicit object parameter, sans identifier ### {#explicit-object-parameter-no-id}
```
BRevzin: I thought BSI hated this syntax, or am I making this up?

atomgalaxy: They did, but Herb loved it. I figured including it is not a bad idea.

0847_deducing_this/deducing-this.bs
```diff
@@ -1993,4 +2049,9 @@ The authors would like to thank:
 - Chandler Carruth for a lot of feedback and guidance around design issues
 - Graham Heynes, Andrew Bennieston, Jeff Snyder for early feedback regarding the meaning of `this` inside function bodies
 - Amy Worthington, Jackie Chen, Vittorio Romeo, Tristan Brindle, Agustín Bergé, Louis Dionne, and Michael Park for early feedback
-- Guilherme Hartmann for his guidance with the implementation
\ No newline at end of file
+- Guilherme Hartmann for his guidance with the implementation
+- Richard Smith and Jens Weller for help with wording
```
BRevzin: Jens Maurer. Also need to add Hubert Tong.

atomgalaxy: done

0847_deducing_this/deducing-this.bs
BRevzin: Yeah, but I feel like the two trailing type syntaxes are at least viable solutions - even if we don't want them. This one I don't think is even viable (unless we go back in time and change what `this` means - in which case it would retroactively the best solution I think).

atomgalaxy: Yes, Herb suggested exactly that - `this` is a reference and behaves exactly the way `self` does in the explicit parameter syntax. In lambdas it shadows captured `this`, since it's a parameter, so you'd need to refer to captured `this` using an assignment alias.

BRevzin: Ok I'm just going to merge this for now and think about it.

PR #21, opened 2019-05-13
Applied most comments from Casey Carter.

tomaszkam:
The only change left is "value of result".
I got current wording from Daniel, and I am not sure about changing it.


BRevzin:
The html file is generated from https://github.com/BRevzin/cpp_proposals/blob/master/1065_constexpr_invoke/constexpr-invoke.md - but my formatting logic isn't exactly documented anywhere (it's based on markdown but I have no idea how much of what I do is local to my environment and uncommitted...)

tomaszkam:
I have done my best attempt to apply the changes to markdown.

PR #25, opened 2020-01-25
Update the build scripts to use the Makefile.

mpark:


1065_constexpr_invoke/constexpr-invoke.md
```diff
@@ -4,10 +4,10 @@ document: P1065R2
...
 date: today
 author:
-	- name: Tomasz Kamiński
-	  email: <tomaszkam@gmail.com>
-	- name: Barry Revzin
-	  email: <barry.revzin@gmail.com>
+  - name: Tomasz Kamiński
+    email: <tomaszkam@gmail.com>
+  - name: Barry Revzin
+    email: <barry.revzin@gmail.com>
```
mpark: Tabs are illegal characters in YAML. This and other Markdown files.

md/defaults.py
```diff
@@ -0,0 +1,13 @@
...
+import pathlib
+
+print(
+"""\
+css: {datadir}/pandoc.css
+filters: [{datadir}/pandoc.py]
+
+metadata:
+  highlighting:
+    inline-code: cpp\
```
mpark: This allows `code_cpp` to go away.

BRevzin: Is this a new pandoc thing or something that always existed and I just didn't know about?

BRevzin:
Thanks for doing this!

I'm going to go ahead and flip all the Makefiles, since like:
```
brevzin@cpp-swim:~/linux/cpp_proposals/2089_param_constraints$ make
make: Circular param-constraints.pdf <- param-constraints.pdf dependency dropped.
make: Nothing to be done for 'all'.
```
but that's a trivial change on my end, just have to do:
```diff
diff --git a/2089_param_constraints/Makefile b/2089_param_constraints/Makefile
index 1a128ba..d4cd99c 100644
--- a/2089_param_constraints/Makefile
+++ b/2089_param_constraints/Makefile
@@ -1,3 +1,2 @@
-include ../md/mpark-wg21.mk
-
 d2089r0.html : param-constraints.md
+include ../md/mpark-wg21.mk
```
everywhere, and I can do that one.

md/defaults.py
mpark: No, I implemented it yesterday: https://github.com/mpark/wg21/commit/f23a31ba796b88248003d4de809c6412648862e0

PR #34, opened 2021-10-16
Fix typo.

hewillk:


BRevzin:
Heh. That's unfortunate. But the paper was already submitted and adopted, so I'm not going to change it now. Thanks for noting it though.

PR #108, opened 2023-10-09
Initial structure for the paper

daveedvdv:
I wrote some initial introductory material, and outlined a bit of the overall paper structure.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -11,3 +11,91 @@ toc: true
 ---
 
 # Introduction
+
+This is a proposal for a reduced initial set of features to support static reflection in C++.
+Specifically, we are mostly proposing a subset of features suggested in P1240R2:
+
+  - the representation of program elements via constant-expressions producing
+     _reflection values_ — _reflections_ for short — of an opaque type `std::meta::info`,
+  - a _reflection operator_ (prefix `^`) that produces a reflection value for its operand construct,
+  - a number of `consteval` _metafunctions_ to work with reflections (including deriving other reflections), and
+  - constructs called _splicers_ to produce grammatical elements from reflections (e.g., `[:` _refl_ `:]`).
+
+This proposal is not intended to be the end-game as far as reflection and compile-time
+metaprogramming are concerned.  Instead, we expect it will be a useful core around which more
+powerful features can be added incrementally over time.  In particular, we believe that most
+or all the remaining features explored in P1240R2 and that code injection
+(along the lines described in P2237R0) are desirable directions to pursue.
+
+Our choice to start with something smaller is primarily motivated by the belief that that
+improves the chances of these facilities making it into the language sooner rather than
+later.
+
+
+## Why a single opaque reflection type?
+
+Perhaps the most common suggestion made regarding the framework outlined in P1240 is to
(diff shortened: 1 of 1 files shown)


2996_reflection/reflection.md
```diff
@@ -11,3 +11,91 @@ toc: true
 ---
 
 # Introduction
+
+This is a proposal for a reduced initial set of features to support static reflection in C++.
+Specifically, we are mostly proposing a subset of features suggested in P1240R2:
```
brevzin, line 16: This will add a reference link:
```suggestion
Specifically, we are mostly proposing a subset of features suggested in [@P1240R2]:
```

2996_reflection/reflection.md
```diff
@@ -11,3 +11,91 @@ toc: true
...
+
+  - the representation of program elements via constant-expressions producing
+     _reflection values_ — _reflections_ for short — of an opaque type `std::meta::info`,
+  - a _reflection operator_ (prefix `^`) that produces a reflection value for its operand construct,
+  - a number of `consteval` _metafunctions_ to work with reflections (including deriving other reflections), and
+  - constructs called _splicers_ to produce grammatical elements from reflections (e.g., `[:` _refl_ `:]`).
```
brevzin, line 22: Can use `$` inside of code blocks to italicize, which means you don't have to leave the code block so this renders nicer:
```suggestion
  - constructs called _splicers_ to produce grammatical elements from reflections (e.g., `[: $refl$ :]`).
```

2996_reflection/reflection.md
```diff
@@ -11,3 +11,91 @@ toc: true
...
+
+This proposal is not intended to be the end-game as far as reflection and compile-time
+metaprogramming are concerned.  Instead, we expect it will be a useful core around which more
+powerful features can be added incrementally over time.  In particular, we believe that most
+or all the remaining features explored in P1240R2 and that code injection
+(along the lines described in P2237R0) are desirable directions to pursue.
```
brevzin, line 28: Ditto reference
```suggestion
(along the lines described in [@P2237R0]) are desirable directions to pursue.
```

2996_reflection/reflection.md
```diff
@@ -11,3 +11,91 @@ toc: true
...
+}
+```
+
+In our initial proposal a value of type `std::meta::info` can represent:
+
+  - an error
+  - any (C++) type and type-alias
+  - any function or member function
+  - any namespace-scope variable or any C++ static data member
+  - any nonstatic data member
```
brevzin, line 88:
```suggestion
  - any non-static data member
```

brevzin (commented):
How do you want to go about reviewing this? Does this work?

Also I think if we're going to have a lot of collaborators, with how diffing works, probably best to split lines as one sentence per line rather than trying to do word wrap on.

daveedvdv:
> How do you want to go about reviewing this? Does this work?

It works.  I'm okay also with you just unilaterally apply what you consider editorial changes.

> 
> Also I think if we're going to have a lot of collaborators, with how diffing works, probably best to split lines as one sentence per line rather than trying to do word wrap on.

I'm editing straight into the GitHub site.  I don't know if there is a way to make that readable...

brevzin (approved):

PR #109, opened 2023-10-10
A bit more text

daveedvdv:


2996_reflection/reflection.md
```diff
@@ -81,21 +110,73 @@ namespace std {
...
   - any namespace-scope variable or any C++ static data member
   - any non-static data member
   - any constant value
   - any template
+  - any namespace
 
 [ (DV) I'm currently excluding local variable representation.  We can change that if we have a use for them that we feel must be in the initial core functionality. ]
 
+
+The type `std::meta::info` is a _scalar_ type with associated namespace `std::meta`.  Nontype template arguments of type `std::meta::info` are permitted.
```
brevzin, line 125: Just for clarity, let's add a parenthetical (or separate sentence, whatever) that means that you can make unqualified calls to the library functions in `std::meta`. Maybe even add that, like, I dunno what's simple...  `std::meta::add_pointer(^int)` and `add_pointer(^int)` are both valid and mean the same thing.

2996_reflection/reflection.md
```diff
@@ -81,21 +110,73 @@ namespace std {
...
+
+Note that _cast-expression_ includes _id-expression_, which in turn can designate templates, member names, etc.
+
+The current proposal requires that the _cast-expression_ be:
+
+  - a _primary-expression_ referring to a function or member function, or
+  - a _primary-expression_ referring to a namespace-scope variable or to a static data member, or
+  - a _primary-expression_ referring to a nonstatic data member, or
+  - a _primary-expression_ referring to a template, or
+  - a constant-expression.
```
brevzin, line 158: I think here let's add some examples of things that you _cannot_ reflect on (in the current proposal). Cause outside of local variable, I can't immediately think of something.

2996_reflection/reflection.md
```diff
@@ -81,21 +110,73 @@ namespace std {
...
+
+In a SFINAE context, a failure to substitute the operand of a reflection operator construct causes that construct to evaluate to an invalid reflection.
 
 ## Splicers (`[:`...`:]`)
 
+A reflection that is not an invalid reflection can be "spliced" into source code using one of several _splicer_ forms:
+
+ - `[: r :]` produces an _expression_ evaluating to the entity or constant value represented by `r`.
+ - `typename[: r :]` produces a _simple-type-specifier_ corresponding to the type represented by `r`.
```
brevzin, line 167: So even in a non-dependent content, I can't write like:
```cpp
using T = [: ^int :];
```
That's invalid, I _must_ write:
```cpp
using T = typename [: ^int :];
```
?

daveedvdv: We can probably allow the same implicit `typename` places as for dependent type names.

brevzin (approved):

PR #110, opened 2023-10-13
Update reflection.md

daveedvdv:
Add some additional examples.
Move examples section to after Proposal section.
Start listing some metafunctions.


2996_reflection/reflection.md
```diff
@@ -271,4 +183,202 @@ std::string name1 = name_of(^S);             // Also okay.
...
+Our next example shows how command-line options could be automatically mapped to an "options" structure.
+For simplicity, we posit the existence of a range-like class type `ProgramArgs` that collects the traditional `(argc, argv)` parameters in a more friendly package.
+This example also uses an expansion statement for simplicity.
+
+::: bq
+```c++
+#include <sstream>
+#include <string>
+#include <meta>
+#include <ProgramArgs.h>
```
brevzin, line 333: I would like to avoid examples that include random headers that readers won't know about. What's `ProgramArgs`? I think it's a lot better here to just use `std::vector<std::string_view>` here.

Also I think the right signature for this is:
```cpp
template <typename Opts>
auto parse_options(int argc, char** argv) -> Opts;
```

Just `std::exit` on failure. That's probably what you want anyway and it makes for a simpler example here.

2996_reflection/reflection.md
```diff
@@ -271,4 +183,202 @@ std::string name1 = name_of(^S);             // Also okay.
...
+
+template<typename Opts> bool parse_options(Opts *opts, ProgramArgs const &args) {
+  using namespace std;
+  using namespace std::meta;
+  bool success = true;
+  for (auto const arg = args.begin(); args != args.end(); ++args) {
+    template for(constexpr auto dm: members_of(^Opts, is_nonstatic_data_member)) {
```
brevzin, line 340: I think it'd be easier to understand the code if we flip the loops:
```cpp
template for (constexpr auto dm : members_of(^Opts, is_nonstatic_data_member)) {
    auto it = ranges::find_if(args, [](string_view arg){ return arg.starts_with("--") && arg.substr(2) == name_of(dm); });
    if (it == args.end()) {
        // not found
    } else if (it + 1 == args.end()) {
        // missing value
    }
    
    // parsing
}
```

2996_reflection/reflection.md
```diff
@@ -271,4 +183,202 @@ std::string name1 = name_of(^S);             // Also okay.
...
+        // Arg is of the form "--word" where "word" is the name of dm.
+        // Move to the option value, but remember the option tag:
+        auto const opt_arg = arg++;
+        if (arg == args.end()) {
+          cerr << "Option " << string(opt_arg) << " is missing a value." << endl;
+          success = false;
+          break;
+        }
+        using T = typename[:type_of(dm):];
+        if constexpr (requires (T d, istringstream is) { is >> d; }) {
```
brevzin, line 351: Should probably be a compile error otherwise?

2996_reflection/reflection.md
```diff
@@ -271,4 +183,202 @@ std::string name1 = name_of(^S);             // Also okay.
...
+          T val;
+          istringstream is(string(*arg));
+          is >> val;
+          opts->[:dm:] = val;
```
brevzin, line 355: If we want to be complete, this part does need to be a runtime error:
```cpp
ispanstream iss(*arg);
if (is >> val) {
    opts.[:dm:] = val;
} else {
    // fail
}
```
or even:
```cpp
ispanstream iss(*arg);
if (is >> opts.[:dm:]; !is) {
    // fail
}
```

2996_reflection/reflection.md
```diff
@@ -271,4 +183,202 @@ std::string name1 = name_of(^S);             // Also okay.
 
 ## Metafunctions
 
+We propose a number of metafunctions declared in namespace `std::meta` to operator on reflection values.
+Adding metafunctions to an implementation is expected to be relatively "easy" compared to implementing the core language features described previously.
+However, despite offering a normal C++ function interface, each on of these relies on "compiler magic" to a significant extent.
```
brevzin, line 188: I think probably want to add a synopsis too. Just nice to have everything in one place.
```cpp
namespace std::meta {
    consteval auto invalid_reflection(string_view message, source_location loc = source_location::current()) -> info;
    consteval auto is_invalid(info) -> bool;
    // ...
}
```

2996_reflection/reflection.md
daveedvdv: I'm fine with that.  Can you apply the change you're thinking of?

2996_reflection/reflection.md
```diff
@@ -271,4 +397,172 @@ std::string name1 = name_of(^S);             // Also okay.
...
+### `invalid_reflection`, `is_invalid`, `diagnose_error`
+
+:::bq
+```c++
+consteval auto invalid_reflection(
+                  std::string_view message,
+                  std::source_location src_loc = std::source_location::current())->info;
+consteval auto is_invalid(info)->bool;
+consteval auto is_invalid(std::span<info>)->bool;
+consteval void diagnose_error(info);
```
brevzin, line 413: I think for these for consistency/clarity present these with the namespace. That way we don't need all the extra `std::`s and since the namespace is _right there_ anyway, should be easy enough to understand (ditto throughout):
```suggestion
namespace std::meta {
  consteval auto invalid_reflection(
                  string_view message,
                  source_location src_loc = source_location::current()) -> info;
  consteval auto is_invalid(info) -> bool;
  consteval auto is_invalid(span<info>) -> bool;
  consteval auto diagnose_error(info) -> void;
}
```
Also, separately - so we have `is_invalid(info)` and... `is_invalid(span<info>)`? I don't think we need the other one. I can already write `std::ranges::any_of(stuff, std::meta::is_invalid)`.

2996_reflection/reflection.md
```diff
@@ -271,4 +397,172 @@ std::string name1 = name_of(^S);             // Also okay.
...
+```
+:::
+
+An invalid reflection represents a potential diagnostic for an erroneous construct.
+Some standard metafunctions will generate such invalid reflections, but user programs can also create them with the `invalid_reflection` metafunction.
+`is_invalid` returns true if it is given an invalid reflection or a span containing at least one invalid reflection.
+Evaluating `diagnose_error` renders a program ill-formed.
+If the given reflection is for an invalid reflection, an implementation is encouraged to render the encapsulated message and source position as part of the diagnostic indicating that the program is ill-formed.
```
brevzin, line 421: Okay so like:
```cpp
std::meta::diagnose_error(my_info);
```
is basically like:
```cpp
if (std::meta::is_invalid(my_info)) {
    std::constexpr_error("Error: my_info is invalid reflection! msg={:?} location={}",
        /* msg? */,
        /* loc? */);
}
```
Right (this is with P2758's `std::constexpr_error`)?

2996_reflection/reflection.md
```diff
@@ -271,4 +397,172 @@ std::string name1 = name_of(^S);             // Also okay.
...
+
+:::bq
+```c++
+consteval auto name_of(info r)->std::string_view;
+consteval auto display_name_of(info r)->std::string_view;
+}
+```
+:::
+
+Given a reflection `r` that designates a declared entity X, `name_of(r)` returns a `string_view` holding the unqualified name of X.
```
brevzin, line 434: So `name_of(^std::vector<int>)` would be `"vector<int>"` (or probably `"vector<int, std::allocator<int>>"`). Is there a way to get `"std::vector<int>"`?

brevzin (commented):
Several typo-level things, couple questions.

2996_reflection/reflection.md
pdimov: Or maybe it will be just `vector`? Names are hard.

2996_reflection/reflection.md
```diff
@@ -271,4 +397,235 @@ std::string name1 = name_of(^S);             // Also okay.
...
+For all other reflections, an empty `string_view` is produced.
+For template instances, the name does not include the template argument list.
```
pdimov, line 436: Yes, `vector` seems the right one.

PR #112, opened 2023-10-16
Fix typos and add a clarification.

daveedvdv:


2996_reflection/reflection.md
```diff
@@ -965,7 +965,7 @@ namespace std::meta {
 ```
 :::
 
-Given a reflection for a template and reflections for template arguments that match that template, `substitute` returns a reflection for the entity obtains by substituting the given arguments in the template.
+Given a reflection for a template and reflections for template arguments that match that template, `substitute` returns a reflection for the entity obtains by substituting the given arguments in the template.  (If the template is a concept template, the substitute result is a constant of type `bool`.)
```
brevzin, line 968: I guess while we're here, there's another grammar issue. Also maybe say that the result is a reflection of a constant?
```suggestion
Given a reflection for a template and reflections for template arguments that match that template, `substitute` returns a reflection for the entity obtained by substituting the given arguments in the template.
If the template is a concept template, the result is a reflection of a constant of type `bool`.
```
Maybe we could add an example like this. Actually, does this equality definitely hold?
```cpp
static_assert(substitute(^std::regular, {^int}) == ^true);
```
At least... this one holds right:
```cpp
static_assert([:substitute(^std::regular, {^int}):] == true);
```

PR #119, opened 2023-11-01
Add compile-time ticket example

daveedvdv:


2996_reflection/reflection.md
```diff
@@ -633,15 +633,52 @@ constexpr auto struct_to_tuple(From const& from) {
...
+class TU_Ticket {
+  template<int N> struct Helper {
+    constexpr int value = N;
+  };
+public:
+  static consteval int next() {
+    // Search for the next incomplete Helper<k>.
+    std::meta::info r;
+    for (int k = 0;; ++k) {
+      auto r = substitute(^Helper, { reflect_value(k) });
```
brevzin, line 665:
```suggestion
      r = substitute(^Helper, { std::meta::reflect_value(k) });
```

2996_reflection/reflection.md
```diff
@@ -633,15 +633,52 @@ constexpr auto struct_to_tuple(From const& from) {
...
+## Compile-Time Ticket Counter
+
+The features proposed here make it a little easier to update a ticket counter at compile time.
+This is not an ideal implementation (we'd prefer direct support for compile-time —-- i.e., `consteval` --- variables), but it shows how compile-time mutable state surfaces in new ways.
+
+::: bq
+```cpp
+class TU_Ticket {
+  template<int N> struct Helper {
+    constexpr int value = N;
```
brevzin, line 658: I think this needs to be:
```suggestion
    static constexpr int value = N;
```
And then, would `value_of<int>(static_data_members_of(r)[0])` actually work? We don't currently have a description of `value_of` for a static data member.

daveedvdv: We say:
> If r is a reflection for a constant-expression or a constant-valued entity of type T, value_of<T>(r) evaluates to that constant value.

that covers constexpr static data members, IMO.  But we do need to add `static_data_members_of` then.

2996_reflection/reflection.md
```diff
@@ -633,15 +633,52 @@ constexpr auto struct_to_tuple(From const& from) {
...
+  };
+public:
+  static consteval int next() {
+    // Search for the next incomplete Helper<k>.
+    std::meta::info r;
+    for (int k = 0;; ++k) {
+      r = substitute(^Helper, { std::meta::reflect_value(k) });
+      if (is_incomplete_type(r)) break;
+    }
+    // Return the value of its member.  Calling nonstatic_data_members_of
```
brevzin, line 668:
```suggestion
    // Return the value of its member.  Calling static_data_members_of
```

brevzin (approved):

PR #132, opened 2024-01-18
Updates to P2996 wording

daveedvdv:
Refine parsing rules for reflection operator a little.
Add constant-evaluation tweaking to deal with side effects.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -1731,6 +1731,7 @@ Change [expr.unary.general]{.sref} paragraph 1 to add productions for the new op
 +    ^ ::
 +    ^ $namespace-name$
 +    ^ $nested-name-specifier$@~opt~@ $template-name$
++    ^ $nested-name-specifier$@~opt~@ $concept-name$
 +    ^ $type-id$
 +    ^ $cast-expression$
 ```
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -1766,6 +1792,59 @@ constexpr auto r = ^std::vector;
...
 
+
+### [expr.const] Constant Expressions
+
+Add a new paragraph after the definition of _manifestly constant-evaluated_ [expr.const]/20:
+
+:::bq
+:::addu
+
+An expression or conversion is _plainly constant-evaluated_ if it is:
```
brevzin, line 1803: I have a bunch of formatting-related comments that are probably easier for me to apply on my own and show you the difference.

Why do we need this other term that seems almost identical to manifestly constant evaluated? I can't actually even tell the distinction between the two.

daveedvdv: One of the tricky things here is that manifestly constant-evaluated can require tentative evaluation.  For example you could imagine a constexpr function g() that calls define_class successfully, then throws in an if-consteval block.  If that function returns an `int` and you write `int x = g();` in namespace scope, it's not manifestly constant-evaluated, but we've already evaluated the define_class call and triggered its side effect.

Similarly, in SFINAE contexts a failure to evaluate can just cause a candidate to be discarded and we wouldn't want that to have side effects.  So plainly constant-evaluated are places where we're sure that the evaluation must succeed.

PR #139, opened 2024-02-04
Add 'reflect_invoke' metafunction and motivating example.

katzdm:
I've discussed this one for a while with @daveedvdv , and we've both had success implementing it. He had no objections, suggested that I write up a motivating example, and propose the wording - so this is that :)

SG7 expressed interest in typeful reflection in Kona, so it's possible the example could help win some new fans to Team 2996.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -1007,10 +1007,9 @@ Although we believe a single opaque `std::meta::info` type to be the best and mo
 ```cpp
 // Represents a 'std::meta::info' constrained by a predicate.
 template <std::meta::info Pred>
-  requires (type_of(reflect_invoke(Pred, {reflect_value(^int)})) == ^bool)
+  requires (type_of(^([:Pred:](^int))) == ^bool)
 struct metatype {
-  // The underlying 'std::meta::info' value.
-  const std::meta::info value;
+  std::meta::info value;
 
   // Construction is ill-formed unless predicate is satisfied.
   consteval metatype(std::meta::info r) : value(r) {
@@ -1019,29 +1018,30 @@ struct metatype {
   }
 
   // Cast to 'std::meta::info' allows values of this type to be spliced.
-  consteval operator const std::meta::info&() const { return value; }
+  consteval operator std::meta::info() const { return value; }
 
   static consteval bool check(std::meta::info r) { return [:Pred:](r); }
-
-  static consteval metatype<Pred> make(std::meta::info r) { return {r}; }
 };
 
 // Type representing a "failure to match" any known metatypes.
-struct unmatched { static consteval unmatched make(...) { return {}; } };
+struct unmatched {
+  consteval unmatched(std::meta::info) {}
+  static consteval bool check(std::meta::info) { return true; }
+};
 
 // Returns the given reflection "enriched" with a more descriptive type.
 template <typename... Choices>
-constexpr auto enrich = [](std::meta::info r) consteval {
-  std::array checks = {^Choices::check...};
-  std::array ctors = {^Choices::make...};
+consteval std::meta::info enrich(std::meta::info r) {
+  std::array ctors = {members_of(^Choices, std::meta::is_constructor)[0]...,
+                      members_of(^unmatched, std::meta::is_constructor)[0]};
+  std::array checks = {^Choices::check..., ^unmatched::check};
 
-  auto choice = ^unmatched;
-  for (auto [check, make] : std::views::zip(checks, makes))
-    if (value_of<bool>(reflect_invoke(check, {reflect_value(r)}))) {
-      choice = make;
-      break;
-    }
-  return reflect_invoke(choice, {reflect_value(r)});
+  std::meta::info choice;
+  for (auto [check, ctor] : std::views::zip(checks, ctors))
+    if (value_of<bool>(reflect_invoke(check, {reflect_value(r)})))
+      return reflect_invoke(ctor, {reflect_value(r)});
+
+  std::unreachable();
 }
 ```
 :::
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -996,6 +997,86 @@ Thus, the only instantiations of `TU_Ticket::Helper` occur because of the call t
...
 [On Compiler Explorer](https://godbolt.org/z/1vEjW4sTr).
 
+## Emulating typeful reflection
+Although we believe a single opaque `std::meta::info` type to be the best and most scalable foundation for reflection, we acknowledge the desire expressed by SG7 for future support for "typeful reflection". The following demonstrates one possible means of assembling a typeful reflection library, in which different classes of reflections are represented by distinct types, on top of the facilities proposed here.
+
+::: bq
+```cpp
+// Represents a 'std::meta::info' constrained by a predicate.
+template <std::meta::info Pred>
+  requires (type_of(reflect_invoke(Pred, {reflect_value(^int)})) == ^bool)
```
brevzin, line 1007: Why do you need `reflect_invoke` here if `Pred` is already a constant? Can't this just be `decltype([: Pred:](^int))`?

2996_reflection/reflection.md
```diff
@@ -996,6 +997,86 @@ Thus, the only instantiations of `TU_Ticket::Helper` occur because of the call t
...
+
+int main() {
+  constexpr auto enrich = ::enrich<type_t, fn_t>;
+
+  // Demonstration of using 'enrich' to select an overload.
+  Printer::PrintKind([:enrich(^int):]);   // "type"
+  Printer::PrintKind([:enrich(^main):]);  // "function"
+  Printer::PrintKind([:enrich(^3):]);     // "unknown kind"
+}
+```
```
brevzin, line 1066: Missing `:::` here.

2996_reflection/reflection.md
```diff
@@ -996,6 +997,86 @@ Thus, the only instantiations of `TU_Ticket::Helper` occur because of the call t
...
+};
+
+// Type representing a "failure to match" any known metatypes.
+struct unmatched { static consteval unmatched make(...) { return {}; } };
+
+// Returns the given reflection "enriched" with a more descriptive type.
+template <typename... Choices>
+constexpr auto enrich = [](std::meta::info r) consteval {
+  std::array checks = {^Choices::check...};
+  std::array ctors = {^Choices::make...};
```
brevzin, line 1033: Is there a way that we can just use the constructors? Since that's really all the `make` functions are here for.

2996_reflection/reflection.md
```diff
@@ -996,6 +997,86 @@ Thus, the only instantiations of `TU_Ticket::Helper` occur because of the call t
...
+Although we believe a single opaque `std::meta::info` type to be the best and most scalable foundation for reflection, we acknowledge the desire expressed by SG7 for future support for "typeful reflection". The following demonstrates one possible means of assembling a typeful reflection library, in which different classes of reflections are represented by distinct types, on top of the facilities proposed here.
+
+::: bq
+```cpp
+// Represents a 'std::meta::info' constrained by a predicate.
+template <std::meta::info Pred>
+  requires (type_of(reflect_invoke(Pred, {reflect_value(^int)})) == ^bool)
+struct metatype {
+  // The underlying 'std::meta::info' value.
+  const std::meta::info value;
```
brevzin, line 1010:
```suggestion
  std::meta::info value;
```

2996_reflection/reflection.md
```diff
@@ -1660,6 +1744,22 @@ namespace std::meta {
...
+  consteval auto reflect_invoke(info target, span<info const> args) -> info;
+}
+```
+:::
+
+This metafunction produces a reflection of the value returned by a call expression.
+
+Letting `F` be the entity reflected by `target`, and `A_0, ..., A_n` be the sequence of entities reflected by the values held by `args`: if the expression `F(A_0, ..., A_N)` is a well-formed constant expression evaluating to a type that is not `void`, and if every value in `args` is a reflection of a constant value, then `reflect_invoke(target, args)` evaluates to a reflection of the constant value `F(A_0, ..., A_N)`.
+
+For all other invocations, `reflect_invoke(target, args)` is ill-formed.
```
brevzin, line 1761: I think we should put `substitute` and `reflect_invoke` next to each other, since they're pretty similar - substitute is `^[: templ :]<[: args :]...>` and `reflect_invoke` is `^[: f :]([: args :]...)`. 

Should also add an example or two.

2996_reflection/reflection.md
```diff
@@ -996,6 +997,86 @@ Thus, the only instantiations of `TU_Ticket::Helper` occur because of the call t
...
+void PrintKind(fn_t) { std::println("function"); }
+void PrintKind(unmatched) { std::println("unknown kind"); }
+
+int main() {
+  constexpr auto enrich = ::enrich<type_t, fn_t>;
+
+  // Demonstration of using 'enrich' to select an overload.
+  Printer::PrintKind([:enrich(^int):]);   // "type"
+  Printer::PrintKind([:enrich(^main):]);  // "function"
+  Printer::PrintKind([:enrich(^3):]);     // "unknown kind"
```
brevzin, line 1064: There's no `Printer` in this example right now.

2996_reflection/reflection.md

katzdm:
@brevzin Addressed all feedback. Did it with constructors, as per your request - I _think_ the existing language of the wording is sufficient to cover that, but let me know if you'd prefer a more explicit mention that calling constructors is fair game.

2996_reflection/reflection.md
```diff
@@ -999,6 +1000,87 @@ Thus, the only instantiations of `TU_Ticket::Helper` occur because of the call t
...
+struct unmatched {
+  consteval unmatched(std::meta::info) {}
+  static consteval bool check(std::meta::info) { return true; }
+};
+
+// Returns the given reflection "enriched" with a more descriptive type.
+template <typename... Choices>
+consteval std::meta::info enrich(std::meta::info r) {
+  std::array ctors = {members_of(^Choices, std::meta::is_constructor)[0]...,
+                      members_of(^unmatched, std::meta::is_constructor)[0]};
```
brevzin, line 1036: Why only taking the first constructor? I guess what you need is the constructor that takes a `std::meta::info`, but this relies on that one being ahead of the copy and move constructors?

katzdm: Well I control the class, and the Standard guarantees that implicitly-declared special member functions are considered to be declared at the closing `}` of the class definition:
https://eel.is/c++draft/class#special

> An implicitly-declared special member function is declared at the closing } of the class-specifier.

And `members_of` guarantees iteration in declaration order, so for the purpose of this example, this seems safe to me.

PR #140, opened 2024-02-08
Add wording for std::meta::info type, plus minor adjustments.

daveedvdv:


2996_reflection/reflection.md
```diff
@@ -1823,6 +1819,48 @@ Change the grammar for `$operator-or-punctuator$` in paragraph 1 of [lex.operato
...
+::: bq
+::: addu
+
+[*]{.pnum} A type is *consteval-only* if it is:
+
+  - `std::meta::info`
+  - an array type with a consteval-only element type
+  - a class type with a consteval-only base class type or consteval-only nonstatic data member type
+  - a pointer or reference to a consteval-only type
+  - a function type with a consteval-only return type or a consteval-only parameter type
```
brevzin, line 1841: I think if we copy the, uh ... structure ... from structural type (which has similar construction) you'd get...

A *consteval-only type* is one of the following:

* `std::meta::info`, or
* a pointer to, reference to, or (possibly multidimensional) array of a consteval-only type, or
* a function type whose return type or any parameter type is a consteval-only type, or
* a class type where the type of any base class or the type of any non-static data member is a consteval-only type. 

In the abstract they're equivalent, but I just think it's mildly better to have the symmetry.

2996_reflection/reflection.md
```diff
@@ -1823,6 +1819,48 @@ Change the grammar for `$operator-or-punctuator$` in paragraph 1 of [lex.operato
...
+
+An object of consteval-only type shall either end its lifetime during the evaluation of a manifestly constant-evaluated expression or conversion ([expr.const]{.sref}), or be a constexpr variable that is not odr-used ([basic.def.odr]{.sref}).
```
brevzin, line 1843: Not pressing now, but should this whole section live in [expr.const] instead?

2996_reflection/reflection.md
```diff
@@ -1823,6 +1819,48 @@ Change the grammar for `$operator-or-punctuator$` in paragraph 1 of [lex.operato
...
+:::
+
+### [basic.fundamental] Fundamental types
+
+Add a new paragraph before the last paragraph of [basic.fundamental]{.sref} as follows:
+
+::: bq
+::: addu
+
+[*]{.pnum} A value of type `std::meta::info` is called a _reflection_ and represent a language element such as a type, a constant value, a nonstatic data member, etc.
```
brevzin, line 1855: Typos
```suggestion
[*]{.pnum} A value of type `std::meta::info` is called a _reflection_ and represents a language element such as a type, a constant value, a non-static data member, etc.
```

2996_reflection/reflection.md
```diff
@@ -2027,6 +2065,8 @@ Add a new subsection in [meta]{.sref} after [type.traits]{.sref}:
 
 ```
 namespace std::meta {
+  using info = decltype(^::);
```
brevzin, line 2068: Surely if the goal is to minimize typing:
```suggestion
  using info = decltype(^0);
```

2996_reflection/reflection.md
```diff
@@ -2235,6 +2275,15 @@ namespace std::meta {
 :::
 :::
 
+### [meta.reflection.info] Reflections
+
+::: bq
+::: addu
+
+[#]{.pnum} The type `std::meta::info` is a synonym for the type produced by the reflection operator ([expr.reflect]{.sref}), and it has the characteristics describes in [basic.types.general]{.sref} and [expr.eq]{.sref}.
```
brevzin, line 2283: I guess we do say something similar for `std::nullptr_t`... I've never seen that sentence before 😄

2996_reflection/reflection.md
daveedvdv: It's not really the goal.  I prefer `decltype(^::)` for emotional reasons.

2996_reflection/reflection.md
daveedvdv: Yeah, I went looking for `nullptr_t` and was also a bit surprised.  It feels a bit circular...

2996_reflection/reflection.md
brevzin: Hehe, okay fair.

2996_reflection/reflection.md
brevzin: Yeah, I kinda feel like we should not add this sentence - we already define what all the characteristics of `info` are... I also kinda feel like we should remove the `nullptr_t` sentence for being similarly useless... 

It's kinda amusing that _because_ we define it this way, this fails:
```cpp
static_assert(^decltype(nullptr) == ^std::nullptr_t);
```
which is kinda weird.

2996_reflection/reflection.md
```diff
@@ -1823,6 +1819,60 @@ Change the grammar for `$operator-or-punctuator$` in paragraph 1 of [lex.operato
...
+::: bq
+[9]{.pnum} Arithmetic types (6.8.2), enumeration types, pointer types, pointer-to-member types (6.8.4),[ `std::meta::info`,]{.addu} `std::nullptr_t`, and cv-qualified (6.8.5) versions of these types are collectively called scalar types.
+:::
+
+Add a new paragraph at the end of [basic.types.general]{.sref} as follows:
+
+::: bq
+::: addu
+
+[*]{.pnum} A *consteval-only* is one of the following:
```
brevzin, line 1835:
```suggestion
[*]{.pnum} A *consteval-only type* is one of the following:
```

2996_reflection/reflection.md
```diff
@@ -1823,6 +1819,60 @@ Change the grammar for `$operator-or-punctuator$` in paragraph 1 of [lex.operato
...
+
+  - `std::meta::info`, or
+  - a pointer or reference to a consteval-only type, or
+  - an (possibly multi-dimensional) array of a consteval-only type, or
+  - a type pointer-to-member of class C of type M where C and/or M is a consteval-only type, or
```
brevzin, line 1840: I think?
```suggestion
  - a pointer-to-member type to a class `C` of type `M` where either `C` or `M` is a consteval-only type, or
```

2996_reflection/reflection.md
```diff
@@ -1823,6 +1819,60 @@ Change the grammar for `$operator-or-punctuator$` in paragraph 1 of [lex.operato
...
+  - a function type with a consteval-only return type or a consteval-only parameter type, or
+  - a class type with a consteval-only base class type or consteval-only nonstatic data member type.
```
brevzin, line 1842:
```suggestion
  - a class type with a consteval-only base class type or consteval-only non-static data member type.
```

2996_reflection/reflection.md
```diff
@@ -1823,6 +1819,60 @@ Change the grammar for `$operator-or-punctuator$` in paragraph 1 of [lex.operato
...
+
+### [basic.lookup.argdep] Argument-dependent name lookup
+
+Add a bullet after the first in paragraph 3 of [basic.lookup.argdep] as follows:
+::: bq
+[3]{.pnum} ... The set of entities is determined in the following way:
+
+- If `T` is a fundamental type, its associated set of entities is empty.
+- [If `T` is `std::meta::info`, its associated set of entities is the singleton containing namespace `std::meta`.]{.addu}
+- If `T` is a class type ...
```
brevzin, line 1873: Adding sub-bullet numbers:
```suggestion
- [3.1]{.pnum} If `T` is a fundamental type, its associated set of entities is empty.
::: addu
- [3.2]{.pnum} If `T` is `std::meta::info`, its associated set of entities is the singleton containing namespace `std::meta`.
:::
- [3.3]{.pnum} If `T` is a class type ...
```

brevzin: Actually is this right? We say:

> The [associated namespaces](https://eel.is/c++draft/basic.lookup.argdep#def:associated_namespaces) for a call are the innermost enclosing non-inline namespaces for its associated entities as well as every element of the inline namespace set ([[namespace.def]](https://eel.is/c++draft/namespace.def)) of those namespaces.

What's the innermost enclosing non-inline namespace of... `std::meta`, isn't that `std`? Should the associated entities of `std::meta::info` be `{std::meta::info}`? Given:

```cpp
namespace std::meta { using info = decltype(^::); }
```
the innermost enclosing non-inline namespace of `std::meta::info` is still, reasonably, `std::meta` right?

katzdm: I think ADL doesn't capture associated namespaces of an alias, hence the need for a special case.

brevzin: I know there's a need for a special case - but as written this isn't correct. I think the associated entity needs to be `std::meta::info` itself, so that the next rule about associated namespaces picks up `std::meta`.

2996_reflection/reflection.md
daveedvdv: Yes, the alias vs aliased entity distinction is a bit of a pain.

2996_reflection/reflection.md
daveedvdv: It does straddle [basic.types] and [expr.const], but I think I prefer it in [basic.types] because it is a type classification.

2996_reflection/reflection.md
daveedvdv: (I made a change along those lines, BTW.)

2996_reflection/reflection.md
daveedvdv: Is the code in the synopsis normative enough to allow dropping that sentence?

2996_reflection/reflection.md
daveedvdv: (I'm still used to the style guide Addison-Wesley required, which prohibited the hyphen after `non` unless pronunciation became ambiguous.)

brevzin: Yeah but we always spell it non-static 🤷

2996_reflection/reflection.md
brevzin: Yeah I think so. I dropped it cause I don't think it adds any information. Also added a missinig "the", otherwise merged.

PR #154, opened 2024-03-10
Various updates for D2996R3.

katzdm:
- Allow direct reflection of deleted functions.
- Allow use of member functions from 'reflect_invoke'.
- Allow use of functions in 'value_of'.
- Various minor edits.


2996_reflection/reflection.md
```diff
@@ -1205,17 +1207,17 @@ using fn_t = metatype<^std::meta::is_function>;
 
 // Example of a function overloaded for different "types" of reflections.
 void PrintKind(type_t) { std::println("type"); }
-void PrintKind(fn_t) { std::println("function"); }
```
brevzin, line 1208: `template_t` needs to be introduced in this example somewhere.

2996_reflection/reflection.md
```diff
@@ -1900,18 +1912,24 @@ namespace std::meta {
...
+If `r` is a reflection of a variable of reference type `T` usable in constant-expressions, then `value_of<T>(r)` evaluates to that reference.
 
-If `r` is a reflection of an enumerator constant of type `E`, `value_of<E>(r)` evaluates to the value of that enumerator.
+If `r` is a reflection of a function, or pointer to a function, of type `R(A_0, ... A_n)`, then `value_of<R(*)(A_0, ..., A_n)>(r)` evaluates to a pointer to that function.
 
-If `r` is a reflection of a non-bit-field non-reference non-static member of type `M` in a class `C`, `value_of<M C::*>(r)` is the pointer-to-member value for that nonstatic member.
+If `r` is a reflection of a non-static member function, or pointer to a non-static member function, and `T` is the type for a pointer to the reflected member function, then `value_of<T>(r)` evaluates to a pointer to the member function.
 
-For other reflection values `r`, `value_of<T>(r)` is ill-formed.
+If `r` is a reflection of a lambda expression, or the closure object of a lambda expression, and the entity reflected by `r` is convertible to `T`, then `value_of<T>(r)` evaluates to `static_cast<T>(r)`.
```
brevzin, line 1926: Er, "convertible"? All of the other rules here are exact. So like `value_of<long>(^0)` is ill-formed, so I think `value_of<void(*)()>(^[]{})` should also be? Or at least, these should be equivalent I think.

2996_reflection/reflection.md
```diff
@@ -1886,9 +1888,19 @@ namespace std::meta {
...
+
+If `F` is a pointer to a non-static member function, `A_0` is a reflection of a value not of a pointer type, and the expression `A_0.*F(A_1, ..., A_N)` is a well-formed constant expression evaluating to a type that is not `void`, then `reflect_invoke(target, args)` evaluates to a reflection of the value `A_0.*F(A_1, ..., A_N)`.
+
+If `F` is a pointer to a non-static member function, `A_0` is a reflection of a value of a pointer type, and the expression `A_0->*F(A_1, ..., A_N)` is a well-formed constant expression evaluating to a type that is not `void`, then `reflect_invoke(target, args)` evaluates to a reflection of the value `A_0->*F(A_1, ..., A_N)`.
+
+If the expression `F(A_0, ..., A_N)` is a well-formed constant expression evaluating to a type that is not `void`, then `reflect_invoke(target, args)` evaluates to a reflection of the constant value `F(A_0, ..., A_N)`.
+
+For all other invocations, `reflect_invoke(target, args)` is not a constant expression.
```
brevzin, line 1903: We should just say that `reflect_invoke(^F, ^{Args...})` evaluates as `std::invoke(F, Args...)`.

2996_reflection/reflection.md
katzdm: Whoops, thanks. Done.

I changed this after a helpful person pointed out that it's illegal to name `main` :-P

2996_reflection/reflection.md
katzdm: Hrm. That would require `value_of<decltype(fn)>(fn)`, which is no good if you're getting a reflection of the lambda as an argument :-\

I don't think it's so bad to have a special case for lambdas. I spelled it as-is to try to allow for generic lambas.

If there's tighter wording we could apply to "convertible", I'm open to suggestions. Also open to revisiting what conversions we allow for other types.

2996_reflection/reflection.md
katzdm: I think we still need special cases for `^S::fn`, since you otherwise would need to spell it `&S::fn`, but that might otherwise work.

2996_reflection/reflection.md
daveedvdv: Don't we still need the sentence that is deleted though?

2996_reflection/reflection.md
katzdm: Yes, it would evaluate to a reflection of a constant value equal to the result of `std::invoke(F, Args...)`.

2996_reflection/reflection.md
katzdm: Didn't mean to delete the "Otherwise" clause; thanks.

Re: lambdas, as per offline discussion, pandas are in, generic pandas are out. :panda_face:

2996_reflection/reflection.md
katzdm: Actually, we also need a special case for where `F` is a function template (which I assume we want to allow).

2996_reflection/reflection.md
brevzin: Wait what? Why not generic lambdas? I _definitely_ dislike _just_ non-generic lambdas.

katzdm: @daveedvdv can confirm, but my understanding is that generic lambdas require emitting AST nodes, where as non-generic lambdas, we can probably pull out the underlying function without needing to. That was the whole "pandas can eat bamboo" remark lol

2996_reflection/reflection.md
katzdm: Oh actually, using `std::invoke` would also rule out constructors. I'd rather keep the language as is for now, and look for opportunities to shorten it later.

2996_reflection/reflection.md
brevzin: I think my strong preference here is simply no conversions allowed (except for qualification conversions, which nobody actually thinks of as conversions anyway).

If we're going to allow _any_ conversions at all, then I would like that decision to be based upon user-facing C++ rules rather than compiler implementation details. Like, as a user, I don't care about AST nodes - I care about understanding the language rules. So like, it would make sense for me if:

* no conversions were allowed, or
* all conversions were allowed, or
* we specifically special-case lambdas as a whole on the premise that otherwise `value_of<T>(lambda_refl)` is basically impossible. But if we special-case lambdas, we should special-case _all_ lambdas not just some lambdas - since the language rule for that conversion is the same for generic and non-generic lambdas, so it should be the same for `value_of` too. 

That's my order of preference right now anyway. I don't like the lambda carve out simply because of the complexity on the user, but I _especially_ dislike the lambda carve out if it doesn't even carve out all lambdas. 

If the starting point here is basically "wait, `value_of` is useless for lambdas unless we do this?" I think it's better to just start with "yes. yes it is." Especially since right now expression reflection is fairly limited.

2996_reflection/reflection.md
brevzin: So the difference between `reflect_invoke(^F, {^Args...})` and `std::invoke(F, Args...)`:

1. you can `reflect_invoke` on a constructor, which you cannot `std::invoke` - there's no actual way to name the constructor. 
2. you can `invoke(&C::mem, std::ref(c))` but you cannot `reflect_invoke(^C::mem, ^std::ref(c))`. 
3. you cannot `reflect_invoke` a function that returns `void`

And possibly, but I'm not sure this is intended:

4. you can `invoke(&C::mem, std::make_unique<C>())` but you cannot `reflect_invoke` on that.

I'd rather describe `reflect_invoke` in terms of a delta from `std::invoke` since people are familiar with that one and this is just prose, cause the important parts are the differences between the two - specifically which ones are on purpose. 

Also why not allow returning `void`?

2996_reflection/reflection.md
katzdm: I've removed the lambda case from `value_of` for the time being. We can circle back later.

2996_reflection/reflection.md
brevzin: Still this one. Guessing just a `using template_t = metatype<^std::meta::is_template>;`?

katzdm: Ugh, thought I fixed this! Thanks, done.

brevzin (approved):

PR #155, opened 2024-03-17
Address some review notes by Christof Meerwald.

daveedvdv:


brevzin:
Merged (but in a different way to actually keep your name on the commit message instead of all of them saying stuff like `Merge branch pr/155`)

PR #157, opened 2024-04-15
Update universal formatter to work with private classes.

katzdm:


brevzin:
I merged this manually and github didn't figure it out.

PR #166, opened 2024-05-14
Allow `if` testability of `std::meta::info` objects

andralex:
Title


andralex:
I should also mention the necessity of hashing and perhaps comparing for inequality. People will be liable to define and use hashtables that map e.g. one type to another:

```c++
std::unordered_map<std::meta::info, std::meta::info> replacement_table;
```

atomgalaxy:
Probably could use the constexpr type comparison paper for the hashing
bit...

brevzin:
I think we decided not to do this one.

PR #167, opened 2024-05-15
Lots of changes. Values and objects, not expressions, not "evaluations".

katzdm:
Also:
- value_of -> extract
- reflect_value -> reflect
- new value_of, is_evaluation metafunctions
- tried to update all uses of "constant value" or "value" to "evaluation" (where appropriate)
- removed support for reflecting cast-expressions
- removed some details of historical discussion (e.g., alternative syntaxes) to avoid further bikeshedding
- editing pass over the entire paper (excluding library wording).

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -1315,7 +1315,7 @@ When the operand is an _id-expression_, the resulting value is a reflection of t
 
 For all other operands, the expression is ill-formed. In a SFINAE context, a failure to substitute the operand of a reflection operator construct causes that construct to not evaluate to constant.
 
-Earlier revisions of this paper allowed for taking the reflection of any _cast-expression_ that could be evaluated as a constant expression, as we believed that a constant expression could be internally "represented" by just capturing the value to which it evaluated. However, the possibility of side effects from constant evaluation (introduced by this very paper) renders this approach infeasible: even a constant expression would have to be evaluated every time it's spliced. It was ultimately decided to defer all support for expression reflection to a future paper.
+Earlier revisions of this paper allowed for taking the reflection of any _cast-expression_ that could be evaluated as a constant expression, as we believed that a constant expression could be internally "represented" by just capturing the value to which it evaluated. However, the possibility of side effects from constant evaluation (introduced by this very paper) renders this approach infeasible: even a constant expression would have to be evaluated every time it's spliced. It was ultimately decided to defer all support for expression reflection, but we intend to introduce it through a future paper using the syntax `^(expr)`.
 
 This paper does, however, support reflections of _values_ and of _objects_ (including subobjects). One way to obtain such reflections is using the `std::meta::reflect_result` metafunction, which returns a reflection of the result of once evaluating its argument. The `std::meta::value_of` metafunction can also be used to obtain a reflection of the value stored by an entity (if the entity is usable in constant expressions). While it's possible to support direct reflection of expression results (e.g., `^fn()`), we aren't convinced that this syntax provided enough value to justify its introduction at this time.
 
@@ -1333,7 +1333,7 @@ That is also not conflicting with the use of the caret as a unary operator becau
 
 Apple also uses the caret in [syntax "blocks"](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html) and unfortunately we believe that does conflict with our proposed use of the caret.
 
-Since the syntax discussions in SG7 landed on the use of the caret, new basic source characters have become available: `@`, `` ` ``{.op}, and `$`{.op}. While we have since discussed some alternatives (e.g., `@` for lifting, `\` and `/` for "raising" and "lowering"), we have grown quite fond of the existing syntax and don't feel that these alternatives offer sufficient value to justify re-painting the bikeshed.
+Since the syntax discussions in SG7 landed on the use of the caret, new basic source characters have become available: `@`, `` ` ``{.op}, and `$`{.op}. While we have since discussed some alternatives (e.g., `@` for lifting, `\` and `/` for "raising" and "lowering"), we have grown quite fond of the existing syntax.
 
 
 ## Splicers (`[:`...`:]`)
@@ -1431,7 +1431,7 @@ For that, the predecessor to this paper, [@P1240R0], proposed an additional form
 Construct the [struct-to-tuple](#converting-a-struct-to-a-tuple) example from above. It was demonstrated using a single splice, but it would be simpler if we had a range splice:
 
 ::: cmptable
-#### With Single Splice
+### With Single Splice
 ```c++
 template <typename T>
 constexpr auto struct_to_tuple(T const& t) {
@@ -1448,7 +1448,7 @@ constexpr auto struct_to_tuple(T const& t) {
 }
 ```
 
-#### With Range Splice
+### With Range Splice
 ```c++
 template <typename T>
 constexpr auto struct_to_tuple(T const& t) {
@@ -1504,6 +1504,16 @@ We propose `[:` and `:]` be single tokens rather than combinations of `[`, `]`, 
 Among others, it simplifies the handling of expressions like `arr[[:refl():]]`.
 On the flip side, it requires a special rule like the one that was made to handle `<::` to leave the meaning of `arr[::N]` unchanged and another one to avoid breaking a (somewhat useless) attribute specifier of the form `[[using ns:]]`.
 
+A syntax that is delimited on the left and right is useful here because spliced expressions may involve lower-precedence operators.
+However, there are other possibilities.
+For example, now that `$`{.op} is available in the basic source character set, we might consider `@[$]{.op}@<$expr$>`.
+This is somewhat natural to those of us that have used systems where `$`{.op} is used to expand placeholders in document templates.  For example:
+::: std
+```c++
+@[$]{.op}@select_type(3) *ptr = nullptr;
+```
+:::
+
 The prefixes `typename` and `template` are only strictly needed in some cases where the operand of the splice is a dependent expression.
 In our proposal, however, we only make `typename` optional in the same contexts where it would be optional for qualified names with dependent name qualifiers.
 That has the advantage to catch unfortunate errors while keeping a single rule and helping human readers parse the intended meaning of otherwise ambiguous constructs.
@@ -1533,7 +1543,7 @@ In our initial proposal a value of type `std::meta::info` can represent:
   - any template
   - any namespace (including the global namespace) or namespace alias
   - any object that is a _permitted result of a constant expression_
-  - any value with _structural type_ (with the caveat that any references or pointers must designate objects that are permitted results of constant expressions)
+  - any value with _structural type_ that is a permitted result of a constant expression
   - the null reflection (when default-constructed)
 
 Notably absent at this time are reflections of expressions. For example, one might wish to walk over the subexpressions of a function call:
@@ -2126,7 +2136,7 @@ namespace std::meta {
 
 If `expr` does not have structural type, then `reflect_result(expr)` fails to be a constant expression.
 
-Otherwise, if `T` is of reference or pointer type, or for each subobject of `expr` having reference or pointer type if `T` is of class type, if the reference or pointer value designates an entity that is not a _permitted result of a constant expression_ ([expr.const]), then `reflect_result(expr)` fails to be a constant expression.
+Otherwise, if `T` is of reference or pointer type, or for each subobject of `expr` having reference or pointer type if `T` is of class type, if the reference or pointer value designates an entity that is not a permitted result ([expr.const]), then `reflect_result(expr)` fails to be a constant expression.
 
 Otherwise, `reflect_result(expr)` produces a reflection of the result of `static_cast<T>(expr)`.
 
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -1504,7 +1504,7 @@ We propose `[:` and `:]` be single tokens rather than combinations of `[`, `]`, 
 Among others, it simplifies the handling of expressions like `arr[[:refl():]]`.
 On the flip side, it requires a special rule like the one that was made to handle `<::` to leave the meaning of `arr[::N]` unchanged and another one to avoid breaking a (somewhat useless) attribute specifier of the form `[[using ns:]]`.
 
-A syntax that is delimited on the left and right is useful here because spliced expressions may involve lower-precedence operators.
+A syntax that is delimited on the left and right is useful here because spliced expressions may involve lower-precedence operators. Additionally, it's important that the left- and right-hand delimiters are different so as to allow nested splices when that comes up. 
 However, there are other possibilities.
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -137,10 +139,11 @@ EDG has an ongoing implementation of this proposal that is currently available o
 
 Additionally, Bloomberg has open sourced a fork of Clang which provides a second implementation of this proposal, also available on Compiler Explorer (again thank you, Matt Godbolt), which can be found here: [https://github.com/bloomberg/clang-p2996](https://github.com/bloomberg/clang-p2996).
 
-Nearly all of the examples below have links to Compiler Explorer demonstrating them in both the EDG and Clang.
+Neither implementation is complete, but all significant features proposed by this paper have been implemented by at least one implementation (including namespace and template splicers). Both implementations have thier "quirks" and continue to evolve alongside this paper.
```
brevzin, line 142: Typo
```suggestion
Neither implementation is complete, but all significant features proposed by this paper have been implemented by at least one implementation (including namespace and template splicers). Both implementations have their "quirks" and continue to evolve alongside this paper.
```

2996_reflection/reflection.md
```diff
@@ -1295,19 +1301,25 @@ The reflection operator produces a reflection value from a grammatical construct
...
+
+The expression `^::` evaluates to a reflection of the global namespace. When the operand is a _namespace-name_ or _type-id_, the resulting value is a reflection of the designated namespace or type.
+
+When the operand is an _id-expression_, the resulting value is a reflection of the designated entity found by lookup. This might be any of:
+
+- a variable, static data member, or structured binding
+- a function or member function
+- a nonstatic data member
+- a template or member template
+- an enumerator
```
brevzin, line 1314: Is the goal to be exhaustive here? So this should include type right?

2996_reflection/reflection.md
```diff
@@ -1295,19 +1301,25 @@ The reflection operator produces a reflection value from a grammatical construct
...
-> |       `^` _cast-expression_
+> |       `^` _id-expression_
+
+The expression `^::` evaluates to a reflection of the global namespace. When the operand is a _namespace-name_ or _type-id_, the resulting value is a reflection of the designated namespace or type.
+
+When the operand is an _id-expression_, the resulting value is a reflection of the designated entity found by lookup. This might be any of:
+
+- a variable, static data member, or structured binding
+- a function or member function
+- a nonstatic data member
```
brevzin, line 1312:
```suggestion
- a non-static data member
```

2996_reflection/reflection.md
```diff
@@ -1295,19 +1301,25 @@ The reflection operator produces a reflection value from a grammatical construct
...
+- a template or member template
+- an enumerator
 
-Note that _cast-expression_ includes _id-expression_, which in turn can designate templates, member names, etc.
+For all other operands, the expression is ill-formed. In a SFINAE context, a failure to substitute the operand of a reflection operator construct causes that construct to not evaluate to constant.
 
-The current proposal requires that the _cast-expression_ be:
+Earlier revisions of this paper allowed for taking the reflection of any _cast-expression_ that could be evaluated as a constant expression, as we believed that a constant expression could be internally "represented" by just capturing the value to which it evaluated. However, the possibility of side effects from constant evaluation (introduced by this very paper) renders this approach infeasible: even a constant expression would have to be evaluated every time it's spliced. It was ultimately decided to defer all support for expression reflection to a future paper.
```
brevzin, line 1318: I think we should note that we intend the syntax for that to be `^(expression)` so that it's clear we're looking ahead?

2996_reflection/reflection.md
```diff
@@ -1418,12 +1426,12 @@ That is, splicing a constructor behaves like a free function that produces an ob
...
 The splicers described above all take a single object of type `std::meta::info` (described in more detail below).
 However, there are many cases where we don't have a single reflection, we have a range of reflections - and we want to splice them all in one go.
-For that, we need a different form of splicer: a range splicer.
+For that, the predecessor to this paper, [@P1240R0], proposed an additional form of splicer: a range splicer.
 
 Construct the [struct-to-tuple](#converting-a-struct-to-a-tuple) example from above. It was demonstrated using a single splice, but it would be simpler if we had a range splice:
 
 ::: cmptable
-### With Single Splice
+#### With Single Splice
```
brevzin, line 1434: No 3 `#`s is correct, that's what `cmptable` is for as a heading.

2996_reflection/reflection.md
```diff
@@ -1534,28 +1529,82 @@ In our initial proposal a value of type `std::meta::info` can represent:
   - any function or member function
   - any variable, static data member, or structured binding
   - any non-static data member
-  - any constant value
+  - any enumerator
   - any template
-  - any namespace
+  - any namespace (including the global namespace) or namespace alias
+  - any object that is a _permitted result of a constant expression_
+  - any value with _structural type_ (with the caveat that any references or pointers must designate objects that are permitted results of constant expressions)
```
brevzin, line 1536: Use the same term:
```suggestion
- any value with _structural type_ that is a permitted result of a constant expression
```

Actually why do we need the requirement on structural? If we end up passing this value into `substitute` then we'll pick up that requirement on substitution, but like... what's wrong with:
```cpp
class C {
    int i; // private
public:
    constexpr int get() const { return i; }
};

constexpr auto r = meta::reflect_result(C{});
static_assert([: r :].get() == 0);
```

2996_reflection/reflection.md
```diff
@@ -2043,39 +2110,54 @@ namespace std::meta {
...
+namespace std::meta {
+  template<typename T> consteval auto reflect_result(T expr) -> info;
+}
+```
+:::
 
-### `value_of<T>`
+If `expr` does not have structural type, then `reflect_result(expr)` fails to be a constant expression.
+
+Otherwise, if `T` is of reference or pointer type, or for each subobject of `expr` having reference or pointer type if `T` is of class type, if the reference or pointer value designates an entity that is not a _permitted result of a constant expression_ ([expr.const]), then `reflect_result(expr)` fails to be a constant expression.
```
brevzin, line 2129: I think you just want to say that `expr` is a permitted result.

2996_reflection/reflection.md
```diff
@@ -2043,39 +2110,54 @@ namespace std::meta {
...
-If `r` is a reflection for a variable of non-reference type `T`, `value_of<T&>(r)` and `value_of<T const&>(r)` are lvalues referring to that variable.
-If the variable is usable in constant expressions [expr.const], `value_of<T>(r)` evaluates to its value.
+If `r` is a reflection for an object of non-reference type `T`, `extract<T&>(r)` and `extract<T const&>(r)` are lvalues referring to that object.
+If the object is usable in constant expressions [expr.const], `extract<T>(r)` evaluates to its value.
 
-If `r` is a reflection for a variable of reference type `T` usable in constant-expressions, `value_of<T>(r)` evaluates to that reference.
+If `r` is a reflection for an object of reference type `T` usable in constant-expressions, `extract<T>(r)` evaluates to that reference.
 
-If `r` is a reflection for a function, or pointer to a function, of type `R(A_0, ... A_n)`, `value_of<R(*)(A_0, ..., A_n)>(r)` evaluates to a pointer to that function.
+If `r` is a reflection for a function of type `R(A_0, ... A_n)`, `extract<R(*)(A_0, ..., A_n)>(r)` evaluates to a pointer to that function.
```
brevzin, line 2150: Just for sanity (and also want to cover `noexcept`, etc.)
```suggestion
If `r` is a reflection for a function of type `F`, `extract<F*>(r)` evaluates to a pointer to that function.
```

2996_reflection/reflection.md
```diff
@@ -2487,21 +2557,16 @@ Add a new subsection of [expr.unary]{.sref} following [expr.delete]{.sref}
...
-Parentheses can be introduced to force the `$cast-expression$` interpretation.
-
-
 [#]{.pnum}
 
 ::: example
 ```
-static_assert(is_type(^int()));    // ^ applies to the type-id "int()"; not the cast "int()"
-static_assert(!is_type(^(int()))); // ^ applies to the the cast-expression "(int())"
+static_assert(is_type(^int()));    // ^ applies to the type-id "int()"
```
brevzin, line 2569: This example is kind of hanging now, since it used to describe the ambiguity. But still useful to have... somewhere. I guess worry about it later.

2996_reflection/reflection.md
```diff
@@ -2517,24 +2582,18 @@ consteval void g(std::meta::info r, X<false> xv) {
...
-The `$cast-expression$` is not evaluated.
+[#]{.pnum} When applied to a `$concept-name$`, the reflection operator produces a reflection for the indicated concept.
 
-* [#.#]{.pnum} If the operand of the reflection operator is an `$id-expression$`, the result is a reflection for the indicated entity.
+[#]{.pnum} When applied to a `$type-id$`, the reflection operator produces a reflection for the indicated type or type alias.
 
-  * [#.#.#]{.pnum} If this `$id-expression$` names an overload set `S`, and if the assignment of `S` to an invented variable of type `const auto` ([dcl.type.auto.deduct]{.sref}) would select a unique candidate function `F` from `S`, the result is a reflection of `F`. Otherwise, the expression `^S` is ill-formed.
+[#]{.pnum} When applied to an `$id-expression$` ([expr.prim.id]{.sref}), the reflection operator produces a reflection of the variable, function, enumerator constant, or nonstatic member designated by the operand.
```
brevzin, line 2593: We hyphenate non-static
```suggestion
[#]{.pnum} When applied to an `$id-expression$` ([expr.prim.id]{.sref}), the reflection operator produces a reflection of the variable, function, enumerator constant, or non-static member designated by the operand.
```

2996_reflection/reflection.md
```diff
@@ -3272,8 +3331,15 @@ consteval bool is_variable_template(info r);
...
+consteval bool is_structured_binding(info r);
+consteval bool is_value(info r);
+```
+[#]{.pnum} *Returns*: `true` if `r` designates a function template, class template, variable template, alias template, concept, structured binding, or value respectively. Otherwise, `false`.
+
+```cpp
+consteval bool is_object(info r);
 ```
-[#]{.pnum} *Returns*: `true` if `r` designates a function template, class template, variable template, alias template, or concept, respectively. Otherwise, `false`.
+[#]{.pnum} *Returns*: `true` if `r` designates an object. Otherwise, `false`.
```
brevzin, line 3342: Why split this one out?

2996_reflection/reflection.md
```diff
@@ -3364,7 +3430,7 @@ template<class... Fs>
 
 [#]{.pnum} *Mandates*: `r` is a reflection designating either a class type or a namespace and `(std::predicate<Fs, info> && ...)` is `true`.
 
-[#]{.pnum} *Returns*: A `vector` containing the reflections of all the direct members `m` of the entity designated by `r` such that `(filters(m) && ...)` is `true`.
+[#]{.pnum} *Returns*: A `vector` containing the reflections of all the direct members `m` of the entity, not including any structured bindings, designated by `r` such that `(filters(m) && ...)` is `true`.
```
brevzin, line 3433:
```suggestion
[#]{.pnum} *Returns*: A `vector` containing the reflections of all the direct members `m` of the entity, excluding any structured bindings, designated by `r` such that `(filters(m) && ...)` is `true`.
```

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
```diff
@@ -1313,22 +1325,18 @@ SG7 eventually agreed upon the prefix `^` operator. The "upward arrow" interpret
...
+Apple also uses the caret in [syntax "blocks"](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html) and unfortunately we believe that does conflict with our proposed use of the caret.
 
-Another option might be the use of the backslash (`\`{.op}).
-It currently has a meaning at the end of a line of source code, but we could still use it as a prefix operator with the constraint that the reflected operand has to start on the same source line.  If we were to opt for that choice, it could make sense to use the slash (`/`{.op}) as a unary operator denoting splicing (see [Splicers](#splicers) below) so that `\`{.op} would correspond to "raise" and `/`{.op} would correspond to "lower".
+Since the syntax discussions in SG7 landed on the use of the caret, new basic source characters have become available: `@`, `` ` ``{.op}, and `$`{.op}. While we have since discussed some alternatives (e.g., `@` for lifting, `\` and `/` for "raising" and "lowering"), we have grown quite fond of the existing syntax and don't feel that these alternatives offer sufficient value to justify re-painting the bikeshed.
```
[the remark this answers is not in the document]
katzdm: I buy that - removed everything following "we have grown fond of the existing syntax."

2996_reflection/reflection.md
katzdm: Thanks, done.

2996_reflection/reflection.md
```diff
@@ -1416,29 +1424,16 @@ constexpr auto struct_to_tuple(T const& t) {
...
-A syntax that is delimited on the left and right is useful here because spliced expressions may involve lower-precedence operators.
-However, there are other possibilities.
-For example, now that `$`{.op} is available in the basic source character set, we might consider `@[$]{.op}@<$expr$>`.
-This is somewhat natural to those of us that have used systems where `$`{.op} is used to expand placeholders in document templates.  For example:
-
-::: std
-```c++
-@[$]{.op}@select_type(3) *ptr = nullptr;
-```
-:::
```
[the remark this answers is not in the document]
katzdm: Fair enough - I'm weakly opposed to re-opening the syntax discussion, but I've reverted the change.

I don't personally feel the need to space around `[:R:]`, but I see your point.

2996_reflection/reflection.md
katzdm: Whatever class of things we allow here, the compiler will have to know how to compare. I think we were trying to rule out custom comparison operators, but I don't think "structural" does that for us anyway.

@daveedvdv Any thoughts on what the right class is here? Aggregate < Structural < Literal, AFAICT

katzdm: (applied the suggested change)

2996_reflection/reflection.md
```diff
@@ -1504,6 +1504,16 @@ We propose `[:` and `:]` be single tokens rather than combinations of `[`, `]`,
 Among others, it simplifies the handling of expressions like `arr[[:refl():]]`.
 On the flip side, it requires a special rule like the one that was made to handle `<::` to leave the meaning of `arr[::N]` unchanged and another one to avoid breaking a (somewhat useless) attribute specifier of the form `[[using ns:]]`.
 
+A syntax that is delimited on the left and right is useful here because spliced expressions may involve lower-precedence operators.
```
brevzin, line 1507: Let's also add this sentence which I thought we had somewhere but we don't. 
```suggestion
A syntax that is delimited on the left and right is useful here because spliced expressions may involve lower-precedence operators. Additionally, it's important that the left- and right-hand delimiters are different so as to allow nested splices when that comes up. 
```

brevzin (approved):

PR #176, opened 2024-06-25
Much of the remaining library wording.

katzdm:
Data layout reflection, extract, reflect_invoke.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -4467,7 +4467,7 @@ consteval size_t offset_of(info r);
 consteval size_t size_of(info r);
 ```
 
-[#]{.pnum} *Mandates*: `r` is a reflection of a type, non-static data member, base class, object, or variable.
+[#]{.pnum} *Mandates*: `r` is a reflection of a type, non-static data member, base class, object, value, or variable.
 
 [#]{.pnum} *Returns* If `r` designates a type `T`, then `sizeof(T)`. Otherwise, `size_of(type_of(r))`.
 
@@ -4483,7 +4483,7 @@ consteval size_t alignment_of(info r);
 consteval size_t bit_offset_of(info r);
 ```
 
-[#]{.pnum} *Mandates*: `r` is a reflection designating a non-static data member.
+[#]{.pnum} *Mandates*: `r` is a reflection designating a non-static data member or a non-virtual base class.
 
 [#]{.pnum} Let `V` be the offset in bits from the beginning of an object of type `parent_of(r)` to the subobject associated with the entity reflected by `r`.
 
@@ -4493,10 +4493,9 @@ consteval size_t bit_offset_of(info r);
 consteval size_t bit_size_of(info r);
 ```
 
-[#]{.pnum} *Mandates*: `r` is a reflection of a type, non-static data member, base class, object, or variable.
+[#]{.pnum} *Mandates*: `r` is a reflection of a type, non-static data member, base class, object, value, or variable.
 
-[#]{.pnum} *Returns* If `r` designates a type, then the size in bits of any object having the reflected type. Otherwise, if `r` designates a variable, base class, object, or non-static data member that is not a bit-field, then `bit_size_of(type_of(r))`. Otherwise, the width of the reflected bit-field.
-
+[#]{.pnum} *Returns* If `r` designates a type, then the size in bits of any object having the reflected type. Otherwise, if `r` reflects a non-static data member that is a bit-field, then the width of the reflected bit-field. Otherwise, `bit_size_of(type_of(r))`.
 :::
 :::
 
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -2437,16 +2439,29 @@ static_assert(dealias(^Y) == ^int);
...
 
+If `r` is a reflection of a variable _usable in constant expressions_, then `object_of(r)` is a reflection of the object designated by the variable. For all other inputs, `object_of(r)` is not a constant expression.
+
+::: std
+```c++
+int x;
+int &y = x;
+
+static_assert(^x != ^y);
+static_assert(object_of(^x) == object_of(^y));
```
brevzin, line 2461: So I think this description isn't quite right because `x` isn't "usable in constant expressions", but `y` is?

But I do think we want these two assertions to work, yeah.

2996_reflection/reflection.md
```diff
@@ -4060,7 +4085,7 @@ consteval bool is_const(info r);
 consteval bool is_volatile(info r);
 ```
 
-[#]{.pnum} *Returns*: `true` if `r` designates a const or volatile type (respectively), a const- or volatile-qualified member function type (respectively), or an object or function with such a type. Otherwise, `false`.
+[#]{.pnum} *Returns*: `true` if `r` designates a const or volatile type (respectively), a const- or volatile-qualified member function type (respectively), or an object, variable, or function with such a type. Otherwise, `false`.
```
brevzin, line 4088:
```cpp
struct C { int const x; };
```
Does `is_const(^C::x)` evaluate to `true` or `false`?

2996_reflection/reflection.md
```diff
@@ -4177,13 +4202,21 @@ consteval info type_of(info r);
 
 [#]{.pnum} *Returns*: A reflection of the type of that entity.  If every declaration of that entity was declared with the same type alias (but not a template parameter substituted by a type alias), the reflection returned is for that alias.  Otherwise, if some declaration of that entity was declared with an alias it is unspecified whether the reflection returned is for that alias or for the type underlying that alias. Otherwise, the reflection returned shall not be a type alias reflection.
 
+```cpp
+consteval info object_of(info r);
+```
+
+[#]{.pnum} *Mandates*: `r` is a reflection designating either an object or a variable usable in constante xpressions ([expr.const]).
```
brevzin, line 4209: Typo
```suggestion
[#]{.pnum} *Mandates*: `r` is a reflection designating either an object or a variable usable in constant expressions ([expr.const]).
```

2996_reflection/reflection.md
```diff
@@ -4423,50 +4456,64 @@ consteval vector<info> accessible_subobjects_of(info target, info from);
...
-consteval size_t alignment_of(info entity);
-consteval size_t bit_offset_of(info entity);
-consteval size_t bit_size_of(info entity);
+consteval size_t offset_of(info r);
 ```
-:::
-:::
 
-### [meta.reflection.result] Expression result reflection {-}
+[#]{.pnum} *Mandates*: `r` is a reflection designating a non-static data member or base class.
```
brevzin, line 4462: Non-virtual base class maybe?

2996_reflection/reflection.md
```diff
@@ -4423,50 +4456,64 @@ consteval vector<info> accessible_subobjects_of(info target, info from);
...
+
+[#]{.pnum} *Returns*: The offset in bytes, from the beginning of any object of type `parent_of(r)`, to the subobject associated with the entity reflected by `r`.
```
brevzin, line 4464: Drop the commas
```suggestion
[#]{.pnum} *Returns*: The offset in bytes from the beginning of any object of type `parent_of(r)` to the subobject associated with the entity reflected by `r`.
```

2996_reflection/reflection.md
```diff
@@ -4423,50 +4456,64 @@ consteval vector<info> accessible_subobjects_of(info target, info from);
...
-template <typename T>
-  consteval info reflect_function(T& expr);
+consteval size_t bit_offset_of(info r);
 ```
 
-[#]{.pnum} *Mandates*: `T` is a function type.
+[#]{.pnum} *Mandates*: `r` is a reflection designating a non-static data member.
 
-[#]{.pnum} *Returns*: `^fn`, where `fn` is the function referenced by `expr`.
+[#]{.pnum} *Returns*: The offset in bits, from the beginning of any object of type `parent_of(r)`, to the subobject associated with the entity reflected by `r`, minus the value of `offset_of(r);`
```
brevzin, line 4488: I'd rather write this as:

```suggestion
[#]{.pnum} Let `V` be the offset in bits from the beginning of an object of type `parent_of(r)` to the subobject associated with the entity reflected by `r`.

[#]{.pnum} *Returns*: `V - offset_of(r)`.
```

Also fixes a typo on any --> an.

I think this makes it more obvious that there's a subtraction going on, otherwise the "minus" is kind of far into the words.

2996_reflection/reflection.md
```diff
@@ -4423,50 +4456,64 @@ consteval vector<info> accessible_subobjects_of(info target, info from);
...
+
+```cpp
+consteval size_t bit_size_of(info r);
+```
+
+[#]{.pnum} *Mandates*: `r` is a reflection of a type, non-static data member, base class, object, or variable.
+
+[#]{.pnum} *Returns* If `r` designates a type, then the size in bits of any object having the reflected type. Otherwise, if `r` designates a variable, base class, or object, then `bit_size_of(type_of(r))`. Otherwise, the size in bits of the subobject associated with the reflected non-static data member within any object of type `parent_of(r)`.
```
brevzin, line 4496: Do we need to say something special about bit-fields?

2996_reflection/reflection.md
```diff
@@ -4518,6 +4565,55 @@ consteval bool test_trait(info templ, R&& arguments);
...
+### [meta.reflection.result] Expression result reflection {-}
+
+::: std
+::: addu
+```cpp
+template <typename T>
+  consteval info reflect_value(T expr);
+```
+
+[#]{.pnum} *Mandates*: `T` is a structural type. `T` is not a reference type. Any subobject of the value computed by `expr` having reference or pointer type designates an entity that is a permitted result of a constant expression.
```
brevzin, line 4577: Can combine these, and add a cross ref.
```suggestion
[#]{.pnum} *Mandates*: `T` is a structural type that is not a reference type. Any subobject of the value computed by `expr` having reference or pointer type designates an entity that is a permitted result of a constant expression ([expr.const]).
```

2996_reflection/reflection.md
```diff
@@ -4518,6 +4565,55 @@ consteval bool test_trait(info templ, R&& arguments);
...
+[#]{.pnum} *Returns*: `^fn`, where `fn` is the function referenced by `expr`.
+
+```cpp
+template <reflection_range R = span<info const>>
+  consteval info reflect_invoke(info target, R&& args);
+template <reflection_range R1 = span<info const>, reflection_range R2 = span<info const>>
+  consteval info reflect_invoke(info target, R1&& tmpl_args, R2&& args);
+```
+
+[#]{.pnum} Let `F` be the entity reflected by `target`, let `Arg0` be the entity reflected by the first element of `args` (if any), let `Args...` be the sequence of entities reflected by the elements of `args` excluding the first, and let `TArgs...` be the sequence of entities or aliases designated by the elements of `tmpl_args`.
```
brevzin, line 4606: "designated" by target?

2996_reflection/reflection.md
```diff
@@ -4518,6 +4565,55 @@ consteval bool test_trait(info templ, R&& arguments);
...
+
+[#]{.pnum} If `F` is a non-member function, a value of pointer to function type, or a value of closure type, then let `INVOKE-EXPR` be the expression `F(Arg0, Args...)`. Otherwise, if `F` is a member function or a value of pointer to member function type, then let `INVOKE-EXPR` be the expression `Arg0.F(Args...)` or `Arg0->F(Args...)` respectively. Otherwise, if `F` is a constructor for a class `C`, then let `INVOKE-EXPR` be the expression `C(Arg0, Args...)` for which only the constructor `F` is considered by overload resolution. Otherwise, if `F` is a non-member function template or a member function template, then let `INVOKE-EXPR` be the expression `F<TArgs...>(Arg0, Args...)` or `Arg0.template F<TArgs...>(Args...)` respectively. Otherwise, if `F` is a constructor template, then let `INVOKE-EXPR` be the expression `C(Arg0, Args...)` for which only the constructor `F` is considered by overload resolution, and `TArgs...` are inferred as explicit template arguments for `F`.
```
brevzin, line 4608: For the first part, I think can lean on `INVOKE(F, Arg0, Args...)` for the first part? Cause it could also be a pointer to member data and it'd be nice to just not have to invent new things. 

I think also that avoids needing to split the arguments, since you just get `INVOKE(F, Args...)` for the first overload and `F<TArgs...>(Args...)` for the second.

2996_reflection/reflection.md
katzdm: Good call - What I want is "variable designating an object having static storage duration".

2996_reflection/reflection.md
katzdm: Yeah, let's make that work.

katzdm: Also "member function types" aren't a thing, so we should just say "const- or volatile-qualified function type" (which ought to capture abominable functions).

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Good call, done.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: I hoped to capture that in the wording about "non-static data members". Just realized that the _width_ of a bit-field is a formal term, so let me state more explicitly in terms of that. Done.

2996_reflection/reflection.md
katzdm: Keeping as is.

2996_reflection/reflection.md
katzdm: Rebased onto INVOKE.

brevzin (approved):

PR #183, opened 2024-07-06
Equality updates, some tweaks, is_enumerator.

katzdm:
- Various tweaks to [expr.eq] for `std::meta::info`.
- Define comparison among `data_member_spec` reflections.
- Reword ODR changes to leverage `data_member_spec` comparison.


2996_reflection/reflection.md
```diff
@@ -3386,14 +3393,17 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
-* [*.#]{.pnum} Otherwise, if one operand is a reflection of a base class specifier and the other is not, then they compare unequal.
-* [*.#]{.pnum} Otherwise, if neither operand is a reflection of a value, then they compare equal if they are reflections of the same entity.
-* [*.#]{.pnum} Otherwise, if one operand is a reflection of a value and the other is not, then they compare unequal.
-* [*.#]{.pnum} Otherwise, if both operands are reflections of values, then they compare equal if and only if the reflected values are _template-argument-equivalent_ ([temp.type]{.sref}).
-* [*.#]{.pnum} Otherwise the result is unspecified. [For example, if the operands are results of calls to `data_member_spec`]{.note}
+* [*.#]{.pnum} Otherwise, if both operands represent entities, then the compare equal if and only if they are reflections of the same entity.
+* [*.#]{.pnum} Otherwise, if one operand represents an entity, then they compare unequal.
+* [*.#]{.pnum} Otherwise, if both operands represent variables, then the compare equal if and only if they are reflections of the same entity.
```
brevzin, line 3398:
```suggestion
* [*.#]{.pnum} Otherwise, if both operands represent variables, then the compare equal if and only if they are reflections of the same variable.
```

2996_reflection/reflection.md
```diff
@@ -3386,14 +3393,17 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
+* [*.#]{.pnum} Otherwise, if one operand represents a variable, then they compare unequal.
+* [*.#]{.pnum} Otherwise, if one operand represents a namespace alias or type alias and the other operand does not represent the same kind of alias, they compare unequal. [A reflection of a type and a reflection of an alias to that same type do not compare equal.]{.note}
```
brevzin, line 3400: What happened to alias template?

2996_reflection/reflection.md
```diff
@@ -3386,14 +3393,17 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
+* [*.#]{.pnum} Otherwise, if both operands represent namespace aliases or type aliases, then they compare equal if and only if their represented aliases share the same name, are declared within the same enclosing scope, and alias the same underlying entity.
+* [*.#]{.pnum} Otherwise, if both operands represent base class specifiers, then they compare equal if and only if they represent the same base class specifier.
+* [*.#]{.pnum} Otherwise, if one operand represents a base class specifier, then they compare unequal.
+* [*.#]{.pnum} Otherwise, if both operands represent values, then they compare equal if and only if the reflected values are _template-argument-equivalent_ ([temp.type]{.sref}) and share the same type.
```
brevzin, line 3404: template-argument-equivalent already implies same type:
```suggestion
* [*.#]{.pnum} Otherwise, if both operands represent values, then they compare equal if and only if the reflected values are _template-argument-equivalent_ ([temp.type]{.sref}).
```

2996_reflection/reflection.md
```diff
@@ -3386,14 +3393,17 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
+* [*.#]{.pnum} Otherwise, if one operand represents a value, then they compare unequal.
+* [*.#]{.pnum} Otherwise, both operands `O@~_1_~@` and `O@~_2_~@` represent descriptions of declarations of non-static data members: Let `C@~_1_~@` and `C@~_2_~@` be invented class types such that each `C@~_k_~@` has a single non-static data member having the properties described by `O@~_k_~@`. The operands compare equal if and only if the data members of `C@~_1_~@` and `C@~_2_~@` would  share the same type, name (if any), `$alignment-specifiers$` (if any), width, and attributes.
```
brevzin, line 3406: What does introducing the class buy us? Can't we just directly say that they compare equal if they share the same type, name, alignment, width, and attributes?

2996_reflection/reflection.md
```diff
@@ -4535,9 +4570,9 @@ consteval info value_of(info r);
 consteval info parent_of(info r);
 ```
 
-[#]{.pnum} *Constant When*: `r` represents a member of either a class or a namespace.
+[#]{.pnum} *Constant When*: `r` represents a variable, structured binding, function, enumerator, class, class member, bit-field, template, namespace, alias of a type or namespace, or base specifier.
 
-[#]{.pnum} *Returns*: A reflection of that entity's immediately enclosing class or namespace.
+[#]{.pnum} *Returns*: A reflection of that entity's immediately enclosing class, function, or namespace.
```
brevzin, line 4575: Did we want to have reflections of scopes? For instance:
```cpp
void f() {
    int x;
    if (true) {
        int y;
    }
}
```
I don't think we want `parent_of(^x) == parent_of(^y)`?

2996_reflection/reflection.md
```diff
@@ -4758,7 +4793,7 @@ consteval size_t size_of(info r);
 consteval size_t alignment_of(info r);
 ```
 
-[#]{.pnum} *Constant When*: `r` is a reflection representing an object, variable, type, non-static data member, or base class.
+[#]{.pnum} *Constant When*: `r` is a reflection representing an object, variable, type, non-static data member that is not a bit-field, or base class.
```
brevzin, line 4796: Bitfields can have alignment, no?

2996_reflection/reflection.md
katzdm: Not quite:

> they are of integral type and their values are the same, or
[(2.2)](https://eel.is/c++draft/temp.type#2.2)
they are of floating-point type and their values are identical, or
[(2.3)](https://eel.is/c++draft/temp.type#2.3)
they are of type std​::​nullptr_t, or
[(2.4)](https://eel.is/c++draft/temp.type#2.4)
they are of enumeration type and their values are the same

The values computed from the literals `3` and `3l` are considered "template-argument-equivalent" ([godbolt](https://godbolt.org/z/Y3s8hYavM)), but I _think_ we want to distinguish them here.

2996_reflection/reflection.md
katzdm: I'm trying to dance around a few concerns here:
1. There's no way to introspect certain properties of a `data_member_spec` (for example: whether it has `no_unique_address` attribute), so it feels a bit odd to directly discuss "which properties" they have.
2. Neither do I want to specify "if it was created with a call to `data_member_spec` having the same arguments...": e.g., arguably `data_member_spec(^int, {.name="m"})` should compare equal to `data_member_spec(^int, {.name=u8"m"})`.

2996_reflection/reflection.md
katzdm: I think we do. I'm not clear that Clang has a natural (persistent) representation of the scope containing `y`; scopes are discarded as they're closed. My understanding is that Clang tends to retain more information than many implementations, so I think it may not be unique in having that problem. We don't have any implementation experience representing block scopes; I don't think we should introduce reflections of them "by fiat".

2996_reflection/reflection.md
katzdm: I don't believe so.

[[dcl.align]](https://eel.is/c++draft/dcl.align#1):
> An [alignment-specifier](https://eel.is/c++draft/dcl.attr.grammar#nt:alignment-specifier) may be applied to a variable or to a class data member, but it **shall not be applied to a bit-field**, a function parameter, or an [exception-declaration](https://eel.is/c++draft/except.pre#nt:exception-declaration) ([[except.handle]](https://eel.is/c++draft/except.handle))[.](https://eel.is/c++draft/dcl.align#1.sentence-1)

[Godbolt](https://godbolt.org/z/Ej3ETr7o6) suggests this is true as well:
```cpp
<source>:5:22: error: invalid application of 'alignof' to bit-field
    5 | static_assert(alignof(S::k) > 0);
```

2996_reflection/reflection.md
katzdm: Ah - As far as I can tell, alias _templates_ are entities, even if their specializations are not - and thus covered by the blanket "entities" clause above.

[[https://eel.is/c++draft/basic.pre#3]](basic.pre):
> An [entity](https://eel.is/c++draft/basic.pre#def:entity) is a value, object, reference, structured binding, function, enumerator, type, class member, bit-field, template, template specialization, namespace, or pack[.](https://eel.is/c++draft/basic.pre#3.sentence-1)

Actually, I might ask about this in CWG on Mattermost - that sentence sort of implies that an alias template specialization is an entity as well.

2996_reflection/reflection.md
katzdm: Good catch, thanks.

2996_reflection/reflection.md
brevzin: Yes quite:

> Two values are [template-argument-equivalent](http://eel.is/c++draft/temp#def:template-argument-equivalent) if **they are of the same type and** [...]

katzdm: Hahahaha whoooops

katzdm: Done, thanks.

2996_reflection/reflection.md
brevzin: Can we just sidestep the whole question and put the aliases (including alias template) before entities? :) I think that'd also make it more obvious what's going on here.

2996_reflection/reflection.md
brevzin: Ok that makes sense. We'll see what CWG says I guess.

2996_reflection/reflection.md
```diff
@@ -3384,16 +3391,15 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
+* [*.#]{.pnum} If one operand is a null reflection value, then they compare equal if and only if the other operand is also a null reflection value.
+* [*.#]{.pnum} Otherwise, if one operand represents a variable, then they compare equal if and only if the other operand represents the same variable.
+* [*.#]{.pnum} Otherwise, if one operand represents a `$template-id$` referring to a specialization of an alias template, then they compare equal if and only if the other operand represents the same `$template-id$` ([temp.type]).
+* [*.#]{.pnum} Otherwise, if one operand represents a namespace alias or a `$typedef-name$`, then they compare equal if and only if the other operand represents a namespace alias or `$typedef-name$` sharing the same name, declared within the same enclosing scope, and aliasing the same underlying entity.
+* [*.#]{.pnum} Otherwise, if one operand represents a value, then they compare equal if and only if the other operand represents a template-argument-equivalent value ([temp.type]{.sref}).
+* [*.#]{.pnum} Otherwise, if one operand represents an object, then they compare equal if and only if the other operand represents the same object.
```
katzdm, line 3399: @brevzin fyi I have a note from St Louis that Jens asked us to separate this case out, since objects may or may not continue to be classified as entities going forward.

brevzin (approved):

PR #184, opened 2024-08-05
Bunch of updates, mostly on the library side.

katzdm:
The biggest change is the replacement of `name_of` and `qualified_name_of` with `identifier_of`, and the addition of certain functions to support that (`operator_of`, `define_static_string`). Also making a few other fixes and extensions that have been discussed. One or two small language fixes, too.

Updated all clang godbolts.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -1400,7 +1400,7 @@ When the operand is an _id-expression_, the resulting value is a reflection of t
 - a variable, static data member, or structured binding
 - a function or member function
 - a non-static data member
-- a template or member template
+- a primary template or primary member template
 - an enumerator
 
 For all other operands, the expression is ill-formed. In a SFINAE context, a failure to substitute the operand of a reflection operator construct causes that construct to not evaluate to constant.
@@ -3153,7 +3153,7 @@ Add a new paragraph before the last paragraph of [basic.fundamental]{.sref} as f
 * a type,
 * a class member,
 * a bit-field,
-* a class template, function template, variable template, alias template, or concept,
+* a primary class template, function template, primary variable template, alias template, or concept,
 * a namespace,
 * an alias of a type or namespace,
 * a base class specifier, or
@@ -4803,11 +4803,23 @@ consteval vector<info> members_of(info r);
 
 [#]{.pnum} *Constant When*: `r` is a reflection representing either a complete class type or a namespace.
 
+[#]{.pnum} A member of a class or namespace is a _representable member_ if it is either
+
+* a class,
+* a `$typedef-name$`,
+* a primary class template, function template, primary variable template, alias template, or concept,
+* a variable or reference,
+* a function,
+* a non-static data member,
+* a namespace, or
+* a namespace alias.
+
 [#]{.pnum} *Effects*: If `dealias(r)` represents a class template specialization with a reachable definition, the specialization is instantiated.
 
-[#]{.pnum} *Returns*: A `vector` containing reflections representing all direct members of the represented entity, excluding any structured bindings. If `r` represents a class `$C$`, then the vector also contains reflections representing all unnamed bit-fields declared within the member-specification of `$C$`.
+[#]{.pnum} *Returns*: A `vector` containing all _representable members_ whose first declaration is directly within a definition of the entity `$E$` represented by `r`. If `$E$` represents a class `$C$`, then the vector also contains reflections representing all unnamed bit-fields declared within the member-specification of `$C$`.
 Non-static data members are indexed in the order in which they are declared, but the order of other kinds of members is unspecified. [Base classes are not members.]{.note}
 
+[#]{.pnum} [Examples of declarations within a class or namespace that are never represented as reflections include: injected class names, partial template specializations, friend declarations, and static assertions.]{.note}
 ```cpp
 consteval vector<info> bases_of(info type);
 ```
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -5196,9 +5196,12 @@ consteval const char* define_static_string(string_view str);
 consteval const char8_t* u8define_static_string(u8string_view str);
 ```
 
-[#]{.pnum} Let `S` be a constexpr variable with static storage duration of type `const char[str.size()]` or `const char8_t[str.size()]` respectively, such that `S[i] == str[i]` for all 0 <= `i` < `str.size()`, and `S[str.size()] == '\0'`.
+[#]{.pnum} Let `S` be a constexpr variable of array type with static storage duration, whose elements are of type `const char` or `const char8_t` respectively, for which there exists some `k >= 0` such that:
 
-[#]{.pnum} *Returns*: `S`
+- `S[k + i] == str[i]` for all 0 <= `i` < `str.size()`, and
+- `S[k + str.size()] == '\0'`.
+
+[#]{.pnum} *Returns*: `&S[k]`
 
 [#]{.pnum} Implementations are encouraged to return the same object whenever the same variant of these functions is called with the same argument.
 
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -33,9 +33,9 @@ Since [@P2996R4]:
 * reduced specification of `is_noexcept`
 * changed `span<info const>` to `initializer_list<info>`
 * removed `test_trait`
-* removed `(u8)name_of` and `(u8)qualified_name_of`; added `(u8)identifier_of`, `operator_of`, `define_static_string`.
+* removed `(u8)name_of` and `(u8)qualified_name_of`; added `(u8)identifier_of`, `operator_of`, `(u8)define_static_string`.
 * renamed `display_name_of` to `display_string_of`
-* adding a number of missing predicates: `is_enumerator`, `is_copy_constructor`, `is_move_constructor`, `is_assignment`, `is_move_assignment`, `is_copy_assignment`, `is_default_constructor`, `has_default_member_initializer`, `is_lvalue_reference_qualified`, `is_rvalue_reference_qualified`, `is_literal_operator(_template)`, `is_conversion_function(_template)`, `is_operator(_template)`)
+* adding a number of missing predicates: `is_enumerator`, `is_copy_constructor`, `is_move_constructor`, `is_assignment`, `is_move_assignment`, `is_copy_assignment`, `is_default_constructor`, `has_default_member_initializer`, `is_lvalue_reference_qualified`, `is_rvalue_reference_qualified`, `is_literal_operator(_template)`, `is_conversion_function(_template)`, `is_operator(_template)`, `is_data_member_spec`
 * changed offset API to be one function that returns a type with named members
 * Tightened constraints on calls to `data_member_spec`, and defined comparison among reflections returned by it.
 * Many wording updates in response to feedback from CWG.
@@ -974,9 +974,8 @@ struct universal_formatter {
 
   template <typename T>
   auto format(T const& t, auto& ctx) const {
-    std::string_view name = has_identifier(^T) ? identifier_of(^T)
-                                               : "(unnamed-type)";
-    auto out = std::format_to(ctx.out(), "{}@{@{", type_label);
+    auto out = std::format_to(ctx.out(), "{}@{@{", has_identifier(^T) ? identifier_of(^T)
+                                                                      : "(unnamed-type)";);
 
     auto delim = [first=true]() mutable {
       if (!first) {
@@ -1037,9 +1036,9 @@ Based on the [@N3980] API:
 ```cpp
 template <typename H, typename T> requires std::is_standard_layout_v<T>
 void hash_append(H& algo, T const& t) {
-  template for (constexpr auto mem : nonstatic_data_members_of(^T))
-    if (!(is_bitfield(mem) && !has_identifier(mem)))
+  template for (constexpr auto mem : nonstatic_data_members_of(^T)) {
       hash_append(algo, t.[:mem:]);
+  }
 }
 ```
 :::
@@ -2174,7 +2173,7 @@ Following much discussion with SG16, we propose #1: The query fails to evaluate 
 
 Earlier revisions of this proposal (and its predecessor, [@P1240R2]) included a metafunction called `name_of`, which we defined to return a `string_view` containing the "name" of the reflected entity. As the paper evolved, it became necessary to sharpen the specification of what this "name" contains. Subsequent revisions (beginning with P2996R2, presented in Tokyo) specified that `name_of` returns the unqualified name, whereas a new `qualified_name_of` would give the fully qualified name.
 
-Most would agree that `qualified_name_of(^size_t)` might reasonably return `"std::size_t"`, or that `qualified_name_of(^std::string::size)` could return `"std::string::size"`. But what about for local variables, or members of local classes? Should inline and anonymous namespaces be rendered as a part of the qualified name? Should we standardize the spelling of such scopes, or leave it implementation defined?
+Most would agree that `qualified_name_of(^size_t)` might reasonably return `"std::size_t"`, or that `qualified_name_of(^std::any::reset)` could return `"std::any::reset"`. But what about for local variables, or members of local classes? Should inline and anonymous namespaces be rendered as a part of the qualified name? Should we standardize the spelling of such scopes, or leave it implementation defined?
 
 The situation is possibly even less clear for unqualified names. Should cv-qualified types be rendered as `const int` or `int const`? Should the type for a function returning a pointer be rendered as `T *(*)()`, `T* (*)()`, or `T * (*)()`? Should such decisions be standardized, or left to implementations? But the real kicker is when one considers non-type template arguments, which can (and do) contain arbitrarily complex values of arbitrary structural types (along with any complete object, or subobject thereof, which has static storage duration).
 
@@ -2400,7 +2399,7 @@ Given a reflection `r` representing a language construct `X` whose declaration i
 The function `u8identifier_of` returns the same identifier but as a `u8string_view`. Note that since all identifiers can be represented as UTF-8 string literals, `u8identifier_of` never fails to be a constant expression because of representability concerns.
 
 Given any reflection `r`, `display_string_of(r)` and `u8display_string_of(r)` return an unspecified non-empty `string_view` and `u8string_view`, respectively.
-Implementations are encouraged to produce text that is helpful in identifying the reflected construct (note: as an exercise, the Clang implementation of this proposal implements a pretty-printing `display_string_of` as a non-intrinsic library function).
+Implementations are encouraged to produce text that is helpful in identifying the reflected construct (note: as an exercise, the Clang implementation of this proposal implements a pretty-printing `display_string_of` [as a non-intrinsic library function](https://github.com/bloomberg/clang-p2996/blob/8ce6449538510a2330f7227f53b40be7671b0b91/libcxx/include/experimental/meta#L2088-L2731)).
 
 Given a reflection `r`, `source_location_of(r)` returns an unspecified `source_location`.
 Implementations are encouraged to produce the correct source location of the item designated by the reflection.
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -937,19 +939,19 @@ struct Clap {
           opts.[:om:] = *cur.initializer;
           continue;
         } else {
-          std::print(stderr, "Missing required option {}\n", name_of(sm));
+          std::print(stderr, "Missing required option {}\n", display_string_of(sm));
```
brevzin, line 942: This one should be fine as `identifier_of` — `sm` is a reflection of a non-static data member. I guess it's the same either way. 

Ditto throughout.

2996_reflection/reflection.md
```diff
@@ -972,7 +974,9 @@ struct universal_formatter {
 
   template <typename T>
   auto format(T const& t, auto& ctx) const {
-    auto out = std::format_to(ctx.out(), "{}@{@{", name_of(^T));
+    std::string_view name = has_identifier(^T) ? identifier_of(^T)
+                                               : "(unnamed-type)";
+    auto out = std::format_to(ctx.out(), "{}@{@{", type_label);
```
brevzin, line 979: Should the variable have been named `type_label`? Would rather just put the whole expression into the call to `format_to`.

2996_reflection/reflection.md
```diff
@@ -989,7 +993,9 @@ struct universal_formatter {
 
     template for (constexpr auto mem : nonstatic_data_members_of(^T)) {
       delim();
-      out = std::format_to(out, ".{}={}", name_of(mem), t.[:mem:]);
+      std::string_view mem_label = has_identifier(mem) ? identifier_of(mem)
+                                                       : "(unnamed-member)";
+      out = std::format_to(out, ".{}={}", mem_label, t.[:mem:]);
```
brevzin, line 998: This one I guess is more interesting since what would you actually want to _do_ in this case? But that's not our job.

2996_reflection/reflection.md
```diff
@@ -1031,9 +1037,9 @@ Based on the [@N3980] API:
 ```cpp
 template <typename H, typename T> requires std::is_standard_layout_v<T>
 void hash_append(H& algo, T const& t) {
-    template for (constexpr auto mem : nonstatic_data_members_of(^T)) {
-        hash_append(algo, t.[:mem:]);
-    }
+  template for (constexpr auto mem : nonstatic_data_members_of(^T))
+    if (!(is_bitfield(mem) && !has_identifier(mem)))
+      hash_append(algo, t.[:mem:]);
```
brevzin, line 1042: I'd rather keep the original instead of making the example more complicated.

2996_reflection/reflection.md
```diff
@@ -2165,6 +2170,21 @@ characters of the result are not representable.  We can then consider multiple o
 
 Following much discussion with SG16, we propose #1: The query fails to evaluate if the identifier cannot be represented in the ordinary string literal encoding.
 
+### Reflecting names
+
+Earlier revisions of this proposal (and its predecessor, [@P1240R2]) included a metafunction called `name_of`, which we defined to return a `string_view` containing the "name" of the reflected entity. As the paper evolved, it became necessary to sharpen the specification of what this "name" contains. Subsequent revisions (beginning with P2996R2, presented in Tokyo) specified that `name_of` returns the unqualified name, whereas a new `qualified_name_of` would give the fully qualified name.
+
+Most would agree that `qualified_name_of(^size_t)` might reasonably return `"std::size_t"`, or that `qualified_name_of(^std::string::size)` could return `"std::string::size"`. But what about for local variables, or members of local classes? Should inline and anonymous namespaces be rendered as a part of the qualified name? Should we standardize the spelling of such scopes, or leave it implementation defined?
```
brevzin, line 2177: Killing me here with picking aliases lol.

2996_reflection/reflection.md
```diff
@@ -2165,6 +2170,21 @@ characters of the result are not representable.  We can then consider multiple o
...
+
+As a nice side-effect, the `identifier_of` model altogether dodges some contentious questions that arose during LEWG discussions in St Louis: Should asking the "name" of an anonymous entity (e.g., anonymous unions) return the empty string, or fail to be a constant expression? Since the C++ grammar requires that an `$identifier$` contain at least one character, the `identifier_of` function never returns an empty string: it is seen that the only possibility is to fail to be a constant expression.
```
brevzin, line 2187: Nicely written throughout 👍

2996_reflection/reflection.md
```diff
@@ -2286,6 +2312,7 @@ namespace std::meta {
   consteval auto is_nonstatic_data_member(info entity) -> bool;
   consteval auto is_static_member(info entity) -> bool;
   consteval auto is_base(info entity) -> bool;
+  consteval auto is_data_member_spec(info r) -> bool;
```
brevzin, line 2315: Note this one in the changelog

2996_reflection/reflection.md
```diff
@@ -2329,6 +2356,10 @@ namespace std::meta {
   template <reflection_range R = initializer_list<info>>
   consteval auto define_class(info type_class, R&&) -> info;
 
+  // @[define_static_string](#define_static_string)@
+  consteval auto define_static_string(string_view str) -> const char *;
+  consteval auto u8define_static_string(u8string_view str) -> const char8_t *;
```
brevzin, line 2361: And this one in the changelog too.

2996_reflection/reflection.md
```diff
@@ -2346,30 +2377,30 @@ namespace std::meta {
...
-For all other reflections, an empty string view is produced.
-For template instances, the name does not include the template argument list.
+The function `u8identifier_of` returns the same identifier but as a `u8string_view`. Note that since all identifiers can be represented as UTF-8 string literals, `u8identifier_of` never fails to be a constant expression because of representability concerns.
 
-Given a reflection `r`, `display_string_of(r)` and `u8display_string_of(r)` return an unspecified non-empty `string_view` and `u8string_view`, respectively.
-Implementations are encouraged to produce text that is helpful in identifying the reflected construct.
+Given any reflection `r`, `display_string_of(r)` and `u8display_string_of(r)` return an unspecified non-empty `string_view` and `u8string_view`, respectively.
+Implementations are encouraged to produce text that is helpful in identifying the reflected construct (note: as an exercise, the Clang implementation of this proposal implements a pretty-printing `display_string_of` as a non-intrinsic library function).
```
brevzin, line 2403: Want to add a link to show off?

2996_reflection/reflection.md
```diff
@@ -2751,6 +2783,45 @@ This allows implementing [variant](#a-simple-variant-type) without having to fur
...
+::: std
+```c++
+namespace std::meta {
+  consteval auto define_static_string(string_view str) -> const char *;
+  consteval auto u8define_static_string(u8string_view str) -> const char8_t *;
+}
+```
+:::
+
+Given a `string_view` or `u8string_view`, returns a pointer to a null-terminated array of characters containing `str` which has static storage duration and is usable in constant expressions.
```
brevzin, line 2796: I don't think we want to require that the _source_ is null-terminated, but we do want to ensure that the _result_ is.

Also not just usable in constant expressions (which string literals are also), we want to make it clear that it's usable as a non-type template parameter. 

Maybe add something like:
```cpp
template <char const* P> struct C { };

char const msg[] = "ok";
C<msg> c1; // works
C<"nope"> c2; // ill-formed
C<define_static_string("yay")> c3; // ok
```

2996_reflection/reflection.md
```diff
@@ -4305,27 +4403,79 @@ namespace std::meta {
...
-consteval string_view name_of(info r);
-consteval u8string_view u8name_of(info r);
-```
 
-[#]{.pnum} *Constant When*: If returning `string_view`, the unqualified name is representable using the ordinary string literal encoding. `r` represents a declared entity.
+[#]{.pnum} This enum class specifies constants used to identify operators that can be overloaded, with the meanings listed in Table 1. The values of the constants are distinct.
+
+<center> Table 1: Enum class `operators` [meta.reflection.operators]</center>
+<table  style="text-align:center">
+<tr><th>[Constant]{.addu}</th><th>[Corresponding operator]{.addu}</th></tr>
```
brevzin, line 4415: Does like... this not work?
```
::: addu
|Constant|Corresponding operator|
|-|-|
|`op_new`|`operator new`|
|`op_delete`|`operator delete`|
...
:::
```

Like as a normal markdown table surrounded by an `addu` fence?

2996_reflection/reflection.md
```diff
@@ -4305,27 +4403,79 @@ namespace std::meta {
...
+<tr><td>[`op_co_await`]{.addu}</td><td>[`operator coawait`]{.addu}</td></tr>
+<tr><td>[`op_parentheses`]{.addu}</td><td>[`operator()`]{.addu}</td></tr>
+<tr><td>[`op_square_brackets`]{.addu}</td><td>[`operator[]`]{.addu}</td></tr>
+<tr><td>[`op_arrow`]{.addu}</td><td>[`operator->`]{.addu}</td></tr>
+<tr><td>[`op_arrow_asterisk`]{.addu}</td><td>[`operator->*`]{.addu}</td></tr>
+<tr><td>[`op_tilde`]{.addu}</td><td>[`operator~`]{.addu}</td></tr>
+<tr><td>[`op_exclamation_mark`]{.addu}</td><td>[`operator!`]{.addu}</td></tr>
+<tr><td>[`op_plus`]{.addu}</td><td>[`operator+`]{.addu}</td></tr>
+<tr><td>[`op_minus`]{.addu}</td><td>[`operator-`]{.addu}</td></tr>
+<tr><td>[`op_asterisk`]{.addu}</td><td>[`operator*`]{.addu}</td></tr>
```
brevzin, line 4429: Can't wait to have discussions about these names. Like should this be `op_star`, etc.

2996_reflection/reflection.md
```diff
@@ -4305,27 +4403,79 @@ namespace std::meta {
...
+
+[#]{.pnum} *Returns*: If `r` represents a literal operator or literal operator template, then the `$ud-suffix$` of the operator or operator template. Otherwise, if `r` represents a variable, entity, or alias of a type or namespace, then the identifier introduced by the the declaration of what is represented by `r`. Otherwise, if `r` represents a base class specifier, then the identifier introduced by the declaration of the type of the base class. Otherwise, if `r` represents a description of the declaration of a non-static data member, then the identifier that would be introduced by the declaration of a data member having the properties represented by `r`.
```
brevzin, line 4478: I'll split this up into bullets later.

2996_reflection/reflection.md
```diff
@@ -4305,27 +4403,79 @@ namespace std::meta {
...
 
-[#]{.pnum} *Returns*: The qualified name of the entity represented by `r`. If this entity has no name, then an empty `string_view` or `u8string_view`, respectively.
+[#]{.pnum} *Constant When*: If `r` represents a function, then when the function is not a constructor, destructor, operator function, or conversion function. Otherwise, if `r` is a function template, then when the function template is not a constructor template, a conversion function template, or an operator function template. Otherwise, if `r` represents a variable, an entity that is not a function or function template, or an alias of a type or namespace, then when the declaration of what is represented by `r` introduces an identifier representable by *E*. Otherwise, if `r` represents a base class specifier for which the base class is a named type, then when the name of that type is an identifier representable by *E*. Otherwise, when `r` represents a description of the declaration of a non-static data member, and the declaration of any data member having the properties represented by `r` would introduce an identifier representable by *E*.
```
brevzin, line 4476: Ok so this:
```suggestion
[#]{.pnum} *Constant When*: `has_identifier(r)` is `true`.
```
Is very nearly correct. It just doesn't cover the representation part. Which makes me wonder if we need `has_{u8,}identifier` to ensure that we can fully check this?

2996_reflection/reflection.md
```diff
@@ -4699,31 +4876,46 @@ consteval vector<info> enumerators_of(info type_enum);
...
+        info target,
+        access_context from = access_context::currrent());
 ```
 
 [#]{.pnum} *Constant When*: `target` is a reflection representing a member or base class specifier of a class.
 
-[#]{.pnum} *Returns*: `true` if the class member or base class specifier represented by `target` can be named within the scope of `from.$context_$`. Otherwise, `false`.
+[#]{.pnum} Let `$C$` be the class for which `target` represents a member or base class specifier.
+
+[#]{.pnum} *Returns*: If `from.$context_$` is the null reflection, then `false`. Otherwise, if `target` represents a class member, then `true` if the member is accessible at all program points within the definition of the entity represented by `from.$context_$` when named in class `$C$` ([class.access]). Otherwise, `true` if the base class represented by `target` is accessible at all program points within the definition of the entity represented by `from.$context_$`. Otherwise, `false`.
```
brevzin, line 4913: Bullets please, since this is already basically naturally there.

2996_reflection/reflection.md
```diff
@@ -4972,6 +5171,22 @@ Defines `class_type` with properties as follows:
...
+::: std
+::: addu
+```cpp
+consteval const char *define_static_string(std::string_view str);
+consteval const char8_t *u8define_static_string(std::u8string_view str);
+```
+
+[#]{.pnum} *Returns*: A pointer to the first element of an array of `char` or `char8_t` objects, respectively, whose length is `str.size() + 1`. The values of the first `str.size()` array subobjects are `str[k]`, and the value of the last array subobject is `0`. The array has static storage duration, and is usable in constant expressions.
+
+[#]{.pnum} [Implementations are encouraged to return the same pointer whenever the same variant of these functions is called with the same argument.]{.note}
```
brevzin, line 5185: Shouldn't be a note, I don't think.

2996_reflection/reflection.md
```diff
@@ -4972,6 +5171,22 @@ Defines `class_type` with properties as follows:
...
+### [meta.reflection.static_string] Static string generation  {-}
+
+::: std
+::: addu
+```cpp
+consteval const char *define_static_string(std::string_view str);
+consteval const char8_t *u8define_static_string(std::u8string_view str);
+```
+
+[#]{.pnum} *Returns*: A pointer to the first element of an array of `char` or `char8_t` objects, respectively, whose length is `str.size() + 1`. The values of the first `str.size()` array subobjects are `str[k]`, and the value of the last array subobject is `0`. The array has static storage duration, and is usable in constant expressions.
```
brevzin, line 5183: I think more like:

* Let `A` be a variable with static storage duration of type `char[str.size()]` or `char8_t[str.size()]` respectively, such that `A[i] == str[i]` for all 0 &le; `i` &lt; `str.size()` and `A[str.size()] == '\0'`. 

* Returns: `A`

2996_reflection/reflection.md
brevzin, line 5185:
```cpp
constexpr char const* a = define_static_string("ghost");
constexpr char const* b = define_static_string("host");
```
Is it allowed for `a + 1 == b`?

2996_reflection/reflection.md
```diff
@@ -4972,6 +5171,22 @@ Defines `class_type` with properties as follows:
 :::
 :::
 
+### [meta.reflection.static_string] Static string generation  {-}
+
+::: std
+::: addu
+```cpp
+consteval const char *define_static_string(std::string_view str);
+consteval const char8_t *u8define_static_string(std::u8string_view str);
```
brevzin, line 5180: Pointer goes on the other side and no `std::`
```suggestion
consteval const char* define_static_string(string_view str);
consteval const char8_t* u8define_static_string(u8string_view str);
```

2996_reflection/reflection.md
katzdm: I would prefer to keep as-is: It happens that `identifier_of(sm)` is fine for non-static data members, but if we were instead outputting a type (as we do below), then it's _not_ necessarily safe - e.g., the type may be an anonymous struct. Rather than draw attention to that distinction, I'm inclined to just blanket-use `display_string_of` everywhere in the example(s) that we're formatting a reflection into text. Let me know if you have a strong preference to use `identifier_of` though, and I can change it.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Reverted 👍

2996_reflection/reflection.md
katzdm: omg lmao - yeah, I'm gonna use `std::any::reset` instead 😂

2996_reflection/reflection.md
katzdm: Thanks, man!

2996_reflection/reflection.md
katzdm: Thanks, done.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Now that I've gotten rid of the `#define STRING_CONSTANT(str)` macro from the implementation, sure! 🤣

2996_reflection/reflection.md
katzdm: > I don't think we want to require that the source is null-terminated, but we do want to ensure that the result is.

I think that's what I've worded, no?

> Also not just usable in constant expressions (which string literals are also), we want to make it clear that it's usable as a non-type template parameter.

> Maybe add something like

Both great ideas; done.

brevzin: > I think that's what I've worded, no?

In the wording section, yes. In the prose here, no.

2996_reflection/reflection.md
katzdm: Haha seems to work 🤣

2996_reflection/reflection.md
katzdm: Gonna be fucking brutal.

2996_reflection/reflection.md
katzdm: It'd be _pretty_ weird if I had some variable named with a UCN, and `has_identifier` returned false.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Thanks, done. Also specified that `A` is constexpr, which I think is needed to ensure that it's "usable in constant expressions".

2996_reflection/reflection.md
katzdm: Done.

katzdm: No, it isn't - But I think that this falls out from the requirement that the returned string is null-terminated.

2996_reflection/reflection.md
katzdm: Hmm - reads that way to me, but I could see how it can be misread. I've rephrased slightly, let me know if it's better.

2996_reflection/reflection.md
brevzin: `"host"` is a suffix of `"ghost"`, with the null terminator. 

The way it's worded right now the answer is no just because we say it's the pointer to the first element of an array. Wonder if that's worth relaxing. Probably not?

katzdm: Oooo great point. Yeah, I guess it can probably always be relaxed later. But here's one way we might say it instead, let me know if you'd like to pursue:

> Let `S` be a constexpr variable of array type with static storage duration, whose elements are of type `const char` or `const char8_t` respectively, for which there exists some `k >= 0` such that:
> - `S[k+i] == str[i]` for all 0 <= `i` < `str.size()`, and
> - `S[k + str.size()] == '\0'`.
>
> _Returns_: `&S[k]`

@brevzin What do you think? Go for it, or leave as is?

brevzin: I like it!

katzdm: Done!

brevzin (approved):

PR #187, opened 2024-08-19
More updates

katzdm:
Staring an R6 with some more updates. Finally passed 100pgs, it's about time.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -1924,7 +1924,7 @@ Still, we are not aware of incompatibilities between our proposal and [@P2758R1]
 
 ### Error-Handling in Reflection
 
-Earlier revisions of this proposal suggested several possible approaches to handling errors in reflection metafunctions. This question arises naturally when considering, for instance, examples like `template_of(^int)`: the argument is a reflection of a type, but that type is not a specialization of a template, so there is no valid reflected template for us to return.
+Earlier revisions of this proposal suggested several possible approaches to handling errors in reflection metafunctions. This question arises naturally when considering, for instance, examples like `template_of(^int)`: the argument is a reflection of a type, but that type is not a specialization of a template, so there is no valid template that we can return.
 
 Some of the possibilities that we have considered include:
 
@@ -2080,6 +2080,26 @@ This paper is proposing that:
   This reflects the actual usage.
 * Meanwhile, `template_arguments_of(^C<int>)` yields `{^int}` while `template_arguments_of(^std::unique_ptr<int>)` yields `{^int, ^std::default_deleter<int>}`.
   This is because `C` has its own template arguments that can be reflected on.
+
+What about when querying the type of an entity?
+
+::: std
+```cpp
+std::string Str;
+const std::string &Ref = Str;
+
+constexpr std::meta::info StrTy = type_of(^Str);
+constexpr std::meta::info RefTy = type_of(^Ref);
+```
+:::
+
+What are `StrTy` and `RefTy`? This question is more difficult. Two distinct issues complicate the answer:
+
+1. Our experience using these facilities has consistently shown that if `StrTy` represents `std::string`, many uses of `StrTy` require writing `dealias(StrTy)` rather than using `StrTy` directly. Failure to do so often yields subtle bugs.
+
+2. While we would like for `RefTy` to represent `const std::string &`, it can only represent `const std::basic_string<char, std::allocator<char>>`. Why? Because since `std::string` is only a "name" for `std::basic_string<char, std::allocator<char>>`, the language provides no semantic answer to what "`const std::string &`" _is_. It is only a source-level "grammatical" construct: A _type-id_. Reflecting type-ids is a brittle path, since it opens questions like whether a reflection of `const int` is the same as a reflection of `int const`. Furthermore, nothing currently requires an implementation to "remember" that the type of `Ref` was "spelled" with the alias `std::string` after parsing it, and we aren't confident that all major implementations do so today.
+
+In light of the above, our position is that `type_of` should never return aliases: That is, `StrTy` represents `std::basic_string<char, std::allocator<char>>`. We believe that it would be desireable to in the future introduce an `aliased_type_of` function capable of returning representations of both `std::string` and `const std::string &` for `StrTy` and `RefTy` respectively - but this requires both discussions with implementers, and likely new wording technology for the Standard. To avoid jeopardizing the goal declared by the title of this paper, we are not proposing such a function at this time.
 
 ### Reflecting source text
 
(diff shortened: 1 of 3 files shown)


2996_reflection/reflection.md
```diff
@@ -2358,9 +2385,14 @@ namespace std::meta {
   consteval auto data_member_spec(info type_class,
                                   data_member_options_t options = {}) -> info;
   template <reflection_range R = initializer_list<info>>
-  consteval auto define_class(info type_class, R&&) -> info;
+    consteval auto define_class(info type_class, R&&) -> info;
+
+  // @[define_static_array](#define_static_array)@
+  template <typename T, input_range<T> R = initializer_list<T>>
+    consteval auto define_static_array(R&& args) -> span<>;
```
brevzin, line 2392: This declaration is broken, what's the intent here?

Also, there's no `#define_static_array` section, no corresponding description of this function, and no wording. Split it up into its own PR?

2996_reflection/reflection.md
```diff
@@ -3122,6 +3156,31 @@ has its linkage determined as follows:
...
+Add a bullet to paragraph 13:
+
+::: std
+
+[13]{.pnum} A declaration `$D$` _names_ an entity `$E$` if
+
+* `$D$` contains a _lambda-expression_ whose closure type is `$E$`,
+* [`$D$` contains a _splice-specifier_ designating `$E$`,]{.addu}
+
+...
```
brevzin, line 3168: When adding bullets, always number them. And them might as well provide the other two bullets for context.

```suggestion
* [13.1]{.pnum} `$D$` contains a _lambda-expression_ whose closure type is `$E$`,
* [13.1+]{.pnum} [`$D$` contains a _splice-specifier_ designating `$E$`,]{.addu}
* [13.2]{.pnum} `$E$` is not a function or function template and `$D$` contains an *id-expression*, *type-specifier*, *nested-name-specifier*, *template-name*, or *concept-name denoting* `$E$`, or
* [13.#]{.pnum} `$E$` is a function or function template and `$D$` contains an expression that names `$E$` ([basic.def.odr]) or an *id-expression* that refers to a set of overloads that contains `$E$`.
```

2996_reflection/reflection.md
```diff
@@ -3122,6 +3156,31 @@ has its linkage determined as follows:
...
+
+Extend the definition of _TU-local_ values and objects to include reflections:
+
+::: std
+
+[16]{.pnum} A value or object is _TU-local_ if either
+
+* [16.1]{.pnum} it is, or is a pointer to, a TU-local function or the object associated with a TU-local variable, [or]{.rm}
+* [[16.2]{.pnum} it is a reflection representing a TU-local entity, or]{.addu}
+* [16.3]{.pnum} it is an object of class or array type and any of its subobjects or any of the objects or functions to which its non-static data members of reference type refer is TU-local and is usable in constant expressions.
```
brevzin, line 3180: Keep the existing numbering (so the middle one becomes `16.1*` or something

2996_reflection/reflection.md
```diff
@@ -26,6 +26,11 @@ tag: constexpr
 
 # Revision History
 
+Since [@P2996R5]:
+
+* removed `op_`-prefixes from `std::meta::operators` enumerators.
```
brevzin, line 31: I'd prefer to back this out.

2996_reflection/reflection.md
```diff
@@ -2074,6 +2081,26 @@ This paper is proposing that:
...
+
+In light of the above, our position is that `type_of` should never return aliases: That is, `StrTy` represents `std::basic_string<char, std::allocator<char>>`. We believe that it would be desireable to in the future introduce an `aliased_type_of` function capable of returning representations of both `std::string` and `const std::string &` for `StrTy` and `RefTy` respectively - but this requires both discussions with implementers, and likely new wording technology for the Standard. To avoid jeopardizing the goal declared by the title of this paper, we are not proposing such a function at this time.
```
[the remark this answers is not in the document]
katzdm: > "desirable" (no extra e).

Gets me every time.

> Also, are we proposing aliased_type_of or not?

No, we are not at this time (for the reasons stated above).

> Like is there no way to get std::string from a reflection of the variable Str?

Correct - though it's possible to compare `type_of(^V) == ^std::string` to know that it's the same underlying type.

> But does anybody actually want those aliases?

Good question.
- `substitute`, I think, _must_ return `some_trait_t<type>`. Consider:

```cpp
template <typename T> class S { };
template <typename T> using A = S<T>;

constexpr auto r = substitute(^A, {^int});
```

Part of the contract of `substitute` is that it only substitutes the template, but does not instantiate it. I think that forming the template-id `S<T>` would go beyond substitution of `A`.

- For `nested_name` (which I don't think we should tackle for P2996), I really have no idea. I don't know whether I find that to be more similar to `type_of` or to `^some_trait<type>::type`.

2996_reflection/reflection.md
katzdm: Oh man, that's super broken lol. Sure, we can do a separate PR.

2996_reflection/reflection.md
```diff
@@ -2358,9 +2385,14 @@ namespace std::meta {
   consteval auto data_member_spec(info type_class,
                                   data_member_options_t options = {}) -> info;
   template <reflection_range R = initializer_list<info>>
-  consteval auto define_class(info type_class, R&&) -> info;
+    consteval auto define_class(info type_class, R&&) -> info;
+
+  // @[define_static_array](#define_static_array)@
+  template <typename T, input_range<T> R = initializer_list<T>>
+    consteval auto define_static_array(R&& args) -> span<>;
+  template <input_range R>
```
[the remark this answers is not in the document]
katzdm: Delorted from this PR instead.

2996_reflection/reflection.md
katzdm: Done, thanks.

2996_reflection/reflection.md
katzdm: Done, used `16.1+` like you did above.

2996_reflection/reflection.md
```diff
@@ -3863,6 +3922,15 @@ Extend *template-argument-equivalent* to handle `std::meta::info`:
 * [2.5]{.pnum} [...]
 :::
 
+### [temp.alias]{.sref} Alias templates {-}
+
+Extend paragraph 2 to enable reflection of alias template specializations.
+
+::: std
+[2]{.pnum} [When]{.rm} [Except when used as the operand of a `$reflect-expression$`,]{.addu} a `$template-id$` refer[ring]{.addu}[s]{.rm} to a specialization of an alias template[, it]{.rm} is equivalent to the associated type obtained by substitution of its `$template-arguments$` for the `$template-parameter$`s in the `$defining-type-id$` of the alias template.
```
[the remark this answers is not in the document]
katzdm: Always ruining my fun.

2996_reflection/reflection.md

brevzin (approved):

PR #188, opened 2024-08-28
Make friends with modules (and some other stuff).

katzdm:
- Removed linkage restrictions put into place at the 11th hour in St. Louis.
- Specified interaction with modules
  - Extend `TU-local` to cover a broader range of reflections.
  - Specify that an expression of reflection type names the entity that it references (for purposes of determining exposures).
  - Define a notion of an _evaluation context_ to help with certain metafunction definitions.
  - Clarify the reflections returned by `members_of` using the new _evaluation context_ notion.
  - Clarify that `define_class` cannot provide a definition for a class attached to another module.
- Align our use of the word "entity" with Davis's theory of it (i.e., variables are entities; values and objects are not).
- Some truly smol nits (e.g., `s/entity/r` in header synopsis).


2996_reflection/reflection.md
```diff
@@ -3158,47 +3143,34 @@ in the *id-expression* of a class member access expression ([expr.ref]). [...]
...
 Extend the definition of _TU-local_ values and objects to include reflections:
 
 ::: std
 
 [16]{.pnum} A value or object is _TU-local_ if either
 
 * [16.1]{.pnum} it is, or is a pointer to, a TU-local function or the object associated with a TU-local variable, [or]{.rm}
-* [[16.1+]{.pnum} it is a reflection representing a TU-local entity, or]{.addu}
+* [[16.1+]{.pnum} it is a value or object of a TU-local type,]{.addu}
+* [[16.1+]{.pnum} it is a reflection representing]{.addu}
```
brevzin, line 3169: nit: these bullets should have different numbers, maybe 1a and 1b.

Also should be able to do something like:
```
::: addu
[16.1a]{.pnum} This,
[16.1b]{.pnum} Or this,
    * [#.#]{.pnum} Don't need to repeat the number here,
    * [#.#]{.pnum Or here
:::
```

As opposed to all the `addu`s? I guess I would expect it to work.

2996_reflection/reflection.md
```diff
@@ -3576,12 +3545,22 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
 ### [expr.const]{.sref} Constant Expressions {-}
 
-Add a new paragraph after the definition of _manifestly constant-evaluated_ [expr.const]{.sref}/20:
+Add a new paragraph after the definition of _potentially constant-evaluated_ [expr.const]{.sref}/21:
 
 ::: std
 ::: addu
 
-[21]{.pnum} An expression or conversion is _plainly constant-evaluated_ if it is:
+[22]{.pnum} The _evaluation context_ is a set of points within the program that determines which declarations are visible to certain expressions used for reflection. During the evaluation of a set of manifestly constant-evaluated expressions `$M$`, let `$E$` be the unique such expression whose result is required for the evaluation of every other expression in `$M$`. The evaluation context comprises the points in the instantiation context of `$E$` ([module.context]).
```
brevzin, line 3553: I don't entirely understand what's going on here. When do we evaluate a "set" of manifestly constant-evaluated expressions? like
```cpp
if constexpr (E1) { ... } 
else if constexpr (E2) { ... }
```
`E1` and `E2` are two different manifestly constant evaluated expressions, what is the unique expression whose result is required for both of these?

katzdm: Re: a "set" of manifestly constant-evaluated expressions:

```cpp
constexpr auto r = [] {
  static constexpr auto s = members_of(^Cls).size();
    // During evaluation of initializer of 's', 'r' is also still being evaluated.
    //
    // The evaluation context would be the program point of 'members_of', not the
    // program point of the lambda.
    
  return s;
};
```

Re: your example above - `E1` is plainly constant evaluated, so `E2` must be parsed in a context in which `E1` has already been evaluated.

(hopefully that helps illustrate what I'm trying to describe - the program point of the expression under evaluation that's "furthest down the stack").

2996_reflection/reflection.md
katzdm: Good catch, and thanks! That worked. Done.

2996_reflection/reflection.md
brevzin: I really have no idea what this means. Definitely could use a few examples? 

Nor do I even understand the `members_of` wording:
> A `vector` containing reflections of all _members-of-representable_ members of the entity represented by `r` that are _members-of-visible_ from one of the points in the evaluation context ([expr.const]).

Which likewise should probably add some examples of cases where some member is not visible.

2996_reflection/reflection.md
```diff
@@ -3158,47 +3145,37 @@ in the *id-expression* of a class member access expression ([expr.ref]). [...]
...
-* [[16.1+]{.pnum} it is a reflection representing a TU-local entity, or]{.addu}
+
+:::addu
+* [16.1a]{.pnum} it is a value or object of a TU-local type,
+* [16.1b]{.pnum} it is a reflection representing
+  * [16.1b.#]{.pnum} a local entity,
+  * [16.1b.#]{.pnum} a TU-local value or object,
+  * [16.1b.#]{.pnum} a `$typedef-name$`, namespace alias, or base specifier introduced by a declaration that is either an exposure or declared at block scope, or
+  * [16.1b.#]{.pnum} an enumerator or a description of a declaration of a nonstatic data member whose type is TU-local, or
+:::
```
DarkArc, line 3178: This is to lift the linkage restriction right?

Could the reflection of an enumerator be considered a TU-local value (and thus we could simplify that here)?

2996_reflection/reflection.md
```diff
@@ -3576,12 +3550,22 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
 
 ### [expr.const]{.sref} Constant Expressions {-}
 
-Add a new paragraph after the definition of _manifestly constant-evaluated_ [expr.const]{.sref}/20:
+Add a new paragraph after the definition of _potentially constant-evaluated_ [expr.const]{.sref}/21:
+
+::: std
+::: addu
+
+[22]{.pnum} The _evaluation context_ is a set of points within the program that determines which declarations are visible to certain expressions used for reflection. During the evaluation of a set of manifestly constant-evaluated expressions `$M$`, let `$E$` be the unique such expression whose result is required for the evaluation of every other expression in `$M$`. The evaluation context comprises the points in the instantiation context of `$E$` ([module.context]).
```
DarkArc, line 3558: I don't know if we need to say this. Following Gaby's feedback, I'm thinking all reachable declarations are presumably usable via reflection ... so this concept of scoping the reflection to what's reachable from the translation unit is perhaps, implicit? Or maybe I don't know exactly what you're going for here?

2996_reflection/reflection.md
```diff
@@ -4993,12 +4988,22 @@ consteval vector<info> members_of(info r);
...
 * a non-static data member,
 * a namespace, or
-* a namespace alias.
+* a namespace alias,
+
+and if its first declaration is within a definition of `$E$`.
 
 [Counterexamples of representable members include: injected class names, partial template specializations, friend declarations, and static assertions.]{.note}
 
+[#]{.pnum} A member `$M$` of a class or namespace is _members-of-visible_ from a point `$P$` if there exists a declaration `$D$` of `$M$` that is reachable from `$P$`, and either `$M$` is not TU-local or `$D$` is declared in the translation unit containing `$P$`.
```
DarkArc, line 4997: > , and either `$M$` is not TU-local or

Do we need to say that? As if it's TU local, wouldn't it be excluded from being part of the module (as otherwise the module would have an exposure)?

2996_reflection/reflection.md
```diff
@@ -4993,12 +4988,22 @@ consteval vector<info> members_of(info r);
...
+
+
 [#]{.pnum} *Effects*: If `dealias(r)` represents a class template specialization with a reachable definition, the specialization is instantiated.
 
+[#]{.pnum} *Returns*: A `vector` containing reflections of all _members-of-representable_ members of the entity represented by `r` that are _members-of-visible_ from one of the points in the evaluation context ([expr.const]).
```
DarkArc, line 5002: > one of the points in the evaluation context

Maybe this is normal but "one of the points" seems ... oddly unspecific

2996_reflection/reflection.md
```diff
@@ -4993,12 +4988,22 @@ consteval vector<info> members_of(info r);
...
+If `$E$` represents a class `$C$`, then the vector also contains reflections representing all unnamed bit-fields declared within the member-specification of `$C$`.
+Non-static data members are indexed in the order in which they are declared, but the order of other kinds of members is unspecified.
```
DarkArc, line 5004: > are indexed in the order in which they are declared

Again maybe this is normal (for core wording) but "are ordered in the order they were declared" or something of that ilk seems better (if possible).

2996_reflection/reflection.md
```diff
@@ -5386,7 +5391,7 @@ consteval bool is_data_member_spec(info r);
 
 [#]{.pnum} *Constant When*: Letting `@$r$~$K$~@` be the `$K$`^th^ reflection value in `mdescrs`,
 
-- `class_type` represents an incomplete class type,
+- `class_type` represents an incomplete class type `$C$`. One of the points in the evaluation context ([expr.const]) belongs to the same module to which `$C$` is attached,
```
DarkArc, line 5394: Similar "one of the points" comment.

2996_reflection/reflection.md
katzdm: These definitions are so convoluted - but I believe TU-local declarations _are_ nevertheless attached to the module. And oddly enough, they are also "reachable" in the strictest sense:

> A declaration _D_ is _reachable from_ a point _P_ if [...]
> - _D_ is not discarded ([module.global.frag]), appears in a translation unit that is reachable from _P_, and does not appear within a private-module-fragment.

So any TU-local definition attached to a named module in a TU visible from _P_ is in fact reachable (TU-local definitions of course don't need to be in the private-module-fragment).

(https://eel.is/c++draft/module.reach#3)

2996_reflection/reflection.md
katzdm: We talked a bit more about this offline - the trouble is that you cannot say that a declaration (member) is simply "reachable", you have to specify where it's reachable _from_ (which is defined using a "program point"). That boils down to two cases:

1. The declaration _D_ is in the same TU as _P_. In that case, it's reachable if _D_ precedes _P_.

2. The declaration _D_ is in a different TU than _P_. Note that the set of "necessarily reachable TUs" is _also_ defined in terms of a program point - so for instance, the declarations from an imported module are not visible prior to the associated `import`-statement (https://eel.is/c++draft/module.reach#1).

In either case, you cannot (formally) discuss reachability without specifying a point - so this is my attempt to formalize the notion that Daveed has called the "point of evaluation".

2996_reflection/reflection.md
katzdm: Are you suggesting that all reflections of enumerators are TU-local values? If so, I'm not sure why we'd want that tighter restriction.

2996_reflection/reflection.md
katzdm: "from a point in the evaluation context" might be a better (and slightly more succinct) way of saying it.

2996_reflection/reflection.md
katzdm: I'm using existing language that we've had for `members_of` and `bases_of`, but I do think I prefer the word "indexed" - it conveys a notion of assigning indices 0, 1, 2, 3, ... in the same order that the members are declared.

2996_reflection/reflection.md
katzdm: Made the same adjustment here; done. The adjustment is minor, but let me know if it reads any better.

2996_reflection/reflection.md
daveedvdv: Is this dropping the definition of "plainly constant evaluated"?

katzdm: > Is this dropping the definition of "plainly constant-evaluated"?

No! I hope not 😅 It's just inserting prior to "plainly constant evaluated", since it's more related to "manifestly constant-evaluated" than it is to "plainly constant-evaluated".

2996_reflection/reflection.md
```diff
@@ -5399,7 +5427,7 @@ consteval bool is_data_member_spec(info r);
 for every `@$r$~$k$~@` in `mdescrs`.
 
 [#]{.pnum} *Effects*:
-Defines `class_type` with properties as follows:
+Produces an injected definition ([expr.const]) of `class_type`, whose locus is immediately after the manifestly constant-evaluated expression whose evaluation is producing the definition, with properties as follows:
```
DarkArc, line 5430: Should this be injected declaration?

katzdm: Since a definition is a declaration, I figured it's probably sufficiently clear to abbreviate "injected declaration that is a definition" as "injected definition" (since this is all so verbose already). What do you think?

DarkArc: I prefer things to be a bit more searchable ... _but_, CWG might disagree with me entirely; so it's probably not worth splitting hairs over.

katzdm: Good point; done.

brevzin (approved):

PR #189, opened 2024-09-23
Remove 'subobjects_of'. A few other minor updates.

katzdm:


2996_reflection/reflection.md
```diff
@@ -4025,6 +4013,26 @@ Adjust paragraph 3 of [temp.arg.general] to not apply to splice template argumen
...
+::: addu
+```cpp
+  constexpr int x = 42;
+  f<[:^int:]>();      // splice-template-argument: calls the first f()
+  f<[:^x:]>();      // splice-template-argument: calls the second f()
+```
+:::
+```cpp
+}
+```
```
brevzin, line 4033: Er? 

You're just adding lines to an existing example right? That's a `diff` code block. 
```
```diff
  template<class T> void f();
  template<int I> void f();

  void g() {
    f<int()>();       // int() is a type-id: call the first f()

+  constexpr int x = 42;
+  f<[:^int:]>();      // splice-template-argument: calls the first f()
+  f<[:^x:]>();      // splice-template-argument: calls the second f()
  }
```

2996_reflection/reflection.md
```diff
@@ -4025,6 +4013,21 @@ Adjust paragraph 3 of [temp.arg.general] to not apply to splice template argumen
...
+template<class T> void f();
+template<int I> void f();
+
+void g() {
+  f<int()>();       // int() is a type-id: call the first f()
+
++ constexpr int x = 42;
++ f<[:^int:]>();    // splice-template-argument: calls the first f()
++ f<[:^x:]>();      // splice-template-argument: calls the second f()
+}
```
brevzin, line 4027: I like adding spaces in everything so that things just line up better (also so the `+`s don't compete with the `}` vertically, makes it stand out a little more):

```suggestion
  template<class T> void f();
  template<int I> void f();

  void g() {
    f<int()>();       // int() is a type-id: call the first f()

+   constexpr int x = 42;
+   f<[:^int:]>();    // splice-template-argument: calls the first f()
+   f<[:^x:]>();      // splice-template-argument: calls the second f()
  }
```

2996_reflection/reflection.md
```diff
@@ -6031,6 +6012,14 @@ references:
...
         day: 19
     URL: https://wg21.link/p3293r0
+  - id: P3293R2
+    citation-label: P3293R2
+    title: "Splicing a base class subobject"
+    author:
+      - family: Peter Dimov
+      - family: Dan Katz
+      - family: Barry Revzin
+      - family: Daveed Vandevoorde
```
brevzin, line 6022: Add the URL here and just write in a date too (just put October 15th, 2024, whatever). And then probably can delete the P3293R0 reference since I think we're not using that anymore. 

I guess actually instead of adding this one, change the other one to be R2. I think while we're here we can remove the explicit reference to P3068R1 since that one's published now too.

brevzin (approved):

PR #197, opened 2024-10-23
A few small changes and fixes.

katzdm:


2996_reflection/reflection.md
```diff
@@ -5570,6 +5571,8 @@ and define an expression `$INVOKE-EXPR$` as follows:
 ::: std
 ::: addu
 
+[1]{.pnum} The class `data_member_options_t` is a consteval-only type ([basic.types.general]), and is not a structural type ([temp.param]).
```
brevzin, line 5574: Need to change the `[1]` after this to a `[#]` (or a `[2]`).

I would comment on that line directly, but github in its infinite wisdom, thinks that there's never any reason to comment more than 3 lines away from a change!

katzdm: Thanks, done.

brevzin (approved):

PR #198, opened 2024-10-28
Some final touch-ups on 'define_class' before LEWG presentation.

katzdm:
- s/unnamed name/unnamed
- Fix a silly mistake in the spec of the "Constant When" for `data_member_spec`
- Make sure reflections of typedefs are allowed for the data member type
- Fail constexpr if two data members will have the same name

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -4822,7 +4822,7 @@ consteval bool has_identifier(info r);
 
 [#]{.pnum} *Returns*:
 
-* [#.#]{.pnum} If `r` represents an entity with an unnamed name, then `false`.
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -5630,6 +5630,7 @@ consteval bool is_data_member_spec(info r);
 - [#.#]{.pnum} `$C$` is incomplete from every point in the evaluation context,
 - [#.#]{.pnum} `is_data_member_spec(@$r$~$K$~@)` is `true` for every `@$r$~$K$~@` in `mdescrs`, and
```
brevzin, line 5631: move the "and" down to the next bullet, which has to end with ", and" instead of "."

brevzin (approved):

PR #200, opened 2024-10-30
Some fixes and removals.

katzdm:
- Remove `define_static_string` and `define_static_array`. Likely to be reintroduced by a follow-up paper.
- Clarify that `source_location_of` gives a source location corresponding to a reachable declaration.
- Clarify the size of a `std::meta::info`.
- Break _plainly constant-evaluated_ in half: _potentially plainly constant-evaluated_ and _plainly constant-evaluated_. The first half stipulates the things that _can_ be plainly constant-evaluated, and the second half filters it down. This structure helps to clarify that a _constant-expression_ resulting from template substitution during argument deduction (or in a requires clause, etc) also fails to be plainly constant-evaluated.
  - Also add a caveat that an immediate invocation appearing within an immediate-escalated function is not plainly constant-evaluated. This also helps the "injected declaration" language to work a bit more cleanly.


brevzin (approved):

brevzin:
Merged this manually in a way that github apparently can't detect.

PR #201, opened 2024-11-04
Rename 'define_class' to 'define_aggregate'.

katzdm:
Also:
* don't fail constexpr if the function called multiple times on the same class, as long as it's called with equal values.
* all non-static data members are public.


2996_reflection/reflection.md
```diff
@@ -2669,7 +2670,7 @@ For other reflection values `r`, `extrace<T>(r)` is ill-formed.
 The function template `extract` may feel similar to splicers, but unlike splicers it does not require its operand to be a constant-expression itself.
 Also unlike splicers, it requires knowledge of the type associated with the entity represented by its operand.
 
-### `data_member_spec`, `define_class`
+### `data_member_spec`, `define_aggregate`
```
brevzin, line 2673: Should add something in here on why we are renaming it to `define_aggregate`

brevzin (approved):

PR #203, opened 2024-11-09
Rewrite a lot of the core wording

katzdm:
- Rewrite and unify the splicer grammars in one place. Overhaul template splicer grammar and handling.
- Rewrite grammar for _reflect-expression_ in terms of lookup (Daveed confirmed). Also fixes some minor wording bugs (e.g., a reflection of a non-type template parameter can yield an object in addition to a value).
- Get rid of `splice-namespace-qualifier`, `splice-template-argument` - it's just `splice-specifier`.
- Rewrite handling of CTAD, deduced placeholder types (i.e., `Concept auto var = value`).
- More careful attention to distinction between names and entities: Lift `*splice-specifier` out of `template-name` and `concept-name`, into `simple-template-id` and `type-constraint`.
- Apply changes discussed for _consteval-only types_ (i.e., immediate-escalation, consteval-erased pointers not being permitted results of constant expressions, etc).
- Splicers, much like `computed-type-specifier`s, are not component names.
- Some QoL changes - prefer the word "construct" in lieu of giant lists of "entity, object, value, base specifier",. etc in some places; introduce the term "splicer" to generically refer to a `splice-specifer`, `splice-expr-template-specifier`, or `splice-type-specifier`.
- Lots of formatting nits (e.g., _word_ vs `$word$`).

And then a lot of other changes spiral out from there.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -3204,29 +3204,29 @@ In all other cases, it is interpreted as a `$splice-expr-template-specifier$`.
 
 ::: example
 ```cpp
-int var {};
+int var;
 struct cls {};
-template<int> int t_var {};
+template<int> int t_var = 0;
 template<int> struct t_cls { };
 template <typename> concept always_true = requires { true; };
 
 template <auto Var, auto Cls, auto TVar, auto TCls, auto Concept>
 void dependent() {
-  { [:Var:] *var; }                        // ok: multiplication of 'var * var'.
+  [:Var:] *var;                      // ok: multiplication of 'var * var'.
 
-  { typename [:Cls:] *var; }               // ok: declaration of 'cls *var'
-  { template [:TVar:]<0> *var; }           // ok: multiplication of 't_var<0> * var'.
-  { template [:TCls:]<0> *var; }           // error: cannot splice type as expression.
-  { typename [:TCls:]<0> *var; }           // ok: declaration of 't_cls<0> *var'.
-  { template typename [:TCls:]<0> *var; }  // also ok.
+  typename [:Cls:] *a;               // ok: declaration of 'cls *var'
+  template [:TVar:]<0> *var;         // ok: multiplication of 't_var<0> * var'.
+  template [:TCls:]<0> *var;         // error: cannot splice type as expression.
+  typename [:TCls:]<0> *b;           // ok: declaration of 't_cls<0> *var'.
+  template typename [:TCls:]<0> *c;  // also ok.
 
-  { template [:Concept:] auto *var = 0; }  // ok: deduced placeholder type.
+  template [:Concept:] auto *d = 0;  // ok: deduced placeholder type.
 };
 
 void non_dependent() {
   dependent<^var, ^cls, ^t_var, ^t_cls, ^always_true>();
 
-  { template [:^^t_cls:]<0> *var; }        // ok in non-dependent context.
+  template [:^t_cls:]<0> *var;  // ok in non-dependent context.
 };
 
 ```
@@ -3273,7 +3273,7 @@ Add a new paragraph restricting `$splice-specifier$`, and renumber accordingly:
 
 ::: std
 ::: addu
-[0]{.pnum} A `$splice-specifier$` of a `$nested-name-specifier$` either is dependent or designates a type, `$typedef-name$`, namespace, or `$namespace-alias$`.
+[0]{.pnum} The `$splice-specifier$` of a `$nested-name-specifier$`, if any, shall either be dependent or designate a class, a `$typedef-name$` denoting a class, a namespace, or a `$namespace-alias$`.
 :::
 
 [1]{.pnum} The component names of a `$qualified-id$` are [...]
@@ -3314,16 +3314,16 @@ $splice-expression$:
    $splice-specifier$
 ```
 
-[#]{.pnum} The `$splice-specifier$` designates either
+[#]{.pnum} The `$splice-specifier$` shall designate either
 
 * [#.#]{.pnum} a value or object,
 * [#.#]{.pnum} a variable,
 * [#.#]{.pnum} a structured binding,
-* [#.#]{.pnum} a function,
+* [#.#]{.pnum} a function other than a constructor or destructor,
 * [#.#]{.pnum} an enumerator, or
 * [#.#]{.pnum} a non-static data member.
 
-The `$splice-specifier$` shall not designate an unnamed bit-field, a constructor, or a destructor.
+[Unnamed bit-fields are not class members ([class.bit]), and cannot be spliced.]{.note}
 
 [#]{.pnum} Let `$E$` be the construct designated by `$splice-specifier$`.
 
@@ -3360,7 +3360,9 @@ Add a production to `$postfix-expression$` for splices in member access expressi
 Modify paragraph 1 to account for splices in member access expressions. Prefer the word "possibly" to "optionally" since `template` cannot precede a `splice-expression`:
 
 ::: std
-[1]{.pnum} A postfix expression followed by a dot `.` or an arrow `->`, [optionally]{.rm} [possibly]{.addu} followed by the keyword template, and then followed by an `$id-expression$` [or a splicer designating a class member]{.addu}, is a postfix expression. [If the keyword `template` is used, the following unqualified name is considered to refer to a template ([temp.names]). If a `$simple-template-id$` results and is followed by a `::`, the `$id-expression$` is a `$qualified-id$`.]{.note}
+[1]{.pnum} A postfix expression followed by a dot `.` or an arrow `->`, [optionally]{.rm} [possibly]{.addu} followed by the keyword `template`, and then followed by an `$id-expression$` [or a splicer designating a class member]{.addu}, is a postfix expression.
+
+[If the keyword `template` is used, the following unqualified name is considered to refer to a template ([temp.names]). If a `$simple-template-id$` results and is followed by a `::`, the `$id-expression$` is a `$qualified-id$`.]{.note}
 
 :::
 
@@ -3442,6 +3444,8 @@ Add a new subsection of [expr.unary]{.sref} following [expr.delete]{.sref}
 ::: std
 ::: addu
 **The Reflection Operator   [expr.reflect]**
+
+[The following grammar avoids use of `$id-expression$`, even though any operand to an `$id-expression$` can be an operand to a `$reflect-expression$`. The reason is that the operand of a `$reflect-expression$` is not necessarily an expression, as it could name a namespace, class template, type, etc.]{.ednote}
 
 ```
 $reflect-expression$:
@@ -3550,17 +3554,32 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
 Modify paragraph 15 to disallow returning non-consteval-only pointers and references to consteval-only objects from constant expressions.
 
 ::: std
-[15]{.pnum} A _constant-expression_ is either a glvalue core constant expression that [\ ]{.addu}
+[15]{.pnum} A _constant expression_ is either a glvalue core constant expression that [\ ]{.addu}
 
 * [#.#]{.pnum} refers to an entity that is a permitted result of a constant expression[, and]{.addu}
 * [[#.#]{.pnum} is of consteval-only type if the entity designated by the expression is of consteval-only type,]{.addu}
 
+  ::: addu
+  ::: note
+  For example, given the function
+  ```cpp
+  consteval Base& fn(const Derived& derived) {
+    return derived;
+  }
+  ```
+
+  where `Derived` is a consteval-only type, if `Base` is not also a consteval-only type then `fn(derived)` is never a constant expression.
+  :::
+  :::
+
 or a prvalue core constant expression whose value satisfies the following constraints:
 
-* [#.#]{.pnum} if the value is an object of class type, each non-static data member of reference type refers to an entity that is a permitted result of a constant expression [and has consteval-only type if the entity has consteval-only type]{.addu},
+* [#.#]{.pnum} if the value is an object of class type, each non-static data member of reference type refers to an entity that is a permitted result of a constant expression [and has consteval-only type if that entity has consteval-only type]{.addu},
 * [#.#]{.pnum} if the value is an object of scalar type, it does not have an indeterminate or erroneous value ([basic.indet]),
-* [#.#]{.pnum} if the value is of pointer type, it is a pointer to an object of static storage duration, a pointer past the end of such an object ([expr.add]), a pointer to a non-immediate function, or a null pointer value,
-* [[#.#]{.pnum} if the value is a pointer to an object of consteval-only type, or a pointer past the end of such an object, then the prvalue is also of consteval-only type,]{.addu}
+* [#.#]{.pnum} if the value is of pointer type, [then:\ ]{.addu}
+  * [#.#.#]{.pnum} it is a pointer to an object of static storage duration[,]{.rm} [or]{.addu} a pointer past the end of such an object ([expr.add]), [such that if the type of that object is a consteval-only type then that pointer type is also a consteval-only type, or]{.addu}
+  * [#.#.#]{.pnum} [it is]{.addu} a pointer to a non-immediate function, or
+  * [#.#.#]{.pnum} [it is]{.addu} a null pointer value,
 * [#.#]{.pnum} if the value is of pointer-to-member-function type, it does not designate an immediate function, and
 * [#.#]{.pnum} if the value is an object of class or array type, each subobject satisfies these constraints for the value.
 
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -143,7 +143,7 @@ Specifically, we are mostly proposing a subset of features suggested in [@P1240R
 
   - the representation of program elements via constant-expressions producing
      _reflection values_ — _reflections_ for short — of an opaque type `std::meta::info`,
-  - a _reflection operator_ (prefix `^`) that produces a reflection value for its operand construct,
+  - a _reflection operator_ (prefix `^`) that computes a reflection value for its operand construct,
   - a number of `consteval` _metafunctions_ to work with reflections (including deriving other reflections), and
   - constructs called _splicers_ to produce grammatical elements from reflections (e.g., `[: $refl$ :]`).
 
@@ -3214,10 +3214,10 @@ template <auto Var, auto Cls, auto TVar, auto TCls, auto Concept>
 void dependent() {
   [:Var:] *var;                      // OK, multiplication of 'var * var'
 
-  typename [:Cls:] *a;               // OK, declaration of 'cls *var'
+  typename [:Cls:] *a;               // OK, declaration of 'cls *a'
   template [:TVar:]<0> *var;         // OK, multiplication of 't_var<0> * var'
   template [:TCls:]<0> *var;         // error: cannot splice type as expression
-  typename [:TCls:]<0> *b;           // OK, declaration of 't_cls<0> *var'
+  typename [:TCls:]<0> *b;           // OK, declaration of 't_cls<0> *b'
   template typename [:TCls:]<0> *c;  // OK
 
   template [:Concept:] auto *d = 0;  // OK, deduced placeholder type
@@ -3443,7 +3443,7 @@ Add a new subsection of [expr.unary]{.sref} following [expr.delete]{.sref}
 
 ::: std
 ::: addu
-**The Reflection Operator   [expr.reflect]**
+**The reflection operator   [expr.reflect]**
 
 [The following grammar avoids use of `$id-expression$`, even though any operand to an `$id-expression$` can be an operand to a `$reflect-expression$`. The reason is that the operand of a `$reflect-expression$` is not necessarily an expression, as it could name a namespace, class template, type, etc.]{.ednote}
 
@@ -3481,11 +3481,11 @@ consteval void g(std::meta::info r, X<false> xv) {
 ```
 :::
 
-[#]{.pnum} A `$reflect-expression$` having the form `^ ::` produces a reflection of the global namespace.
+[#]{.pnum} A `$reflect-expression$` having the form `^ ::` computes a reflection of the global namespace.
 
-[#]{.pnum} A `$reflect-expression$` having the form `^ $unqualified-id$` or `^ $qualified-id$` performs name lookup for the operand following `^` and produces a result as follows:
+[#]{.pnum} A `$reflect-expression$` having the form `^ $unqualified-id$` or `^ $qualified-id$` performs name lookup for the operand following `^` and computes a result as follows:
 
-* [#.#]{.pnum} If the lookup finds an overload set `$S$` such that the assignment of `&$S$` to an invented variable of type `const auto` ([dcl.type.auto.deduct]{.sref}) would select a unique candidate function `$F$` from `$S$`, then the result is a reflection of `$F$`. If the lookup finds any other overload set, the program is ill-formed.
+* [#.#]{.pnum} If the lookup finds an overload set `$S$` such that the initialization of an invented variable of type `const auto` ([dcl.type.auto.deduct]{.sref}) with `&$S$` would select a unique candidate function `$F$` from `$S$`, then the result is a reflection of `$F$`. If the lookup finds any other overload set, the program is ill-formed.
 
 * [#.#]{.pnum} Otherwise, if the lookup finds a `$typedef-name$` or a `$namespace-alias$`, then the result is a reflection of the indicated name.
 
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -3123,6 +3127,22 @@ An expression convertible to a reflection is said to _represent_ the correspondi
 :::
 :::
 
+### [conv.ptr]{.sref} Pointer conversions {-}
+
+Expand paragraph 2 to disallow conversions to `void *` from pointers of consteval-only types.
```
brevzin, line 3132: Uh, what? Why? This effectively prohibits any type erasure involving consteval-only types, even at compile time.

2996_reflection/reflection.md
```diff
@@ -3123,6 +3127,22 @@ An expression convertible to a reflection is said to _represent_ the correspondi
...
+
+::: std
+[2]{.pnum} A prvalue of type "pointer to `$cv$ T`", where `T` is an object type, can be converted to a prvalue of type "pointer to `$cv$ void`". The pointer value ([basic.compound]) is unchanged by this conversion. [If `T` is a consteval-only type ([basic.types.general]), a program that necessitates this conversion is ill-formed.]{.addu}
+
+:::
+
+Expand paragraph 3 to also disallow conversions from pointers of consteval-only type to pointers to base classes that are not of consteval-only type.
```
brevzin, line 3139: And this would prohibit the _other_ way of doing type erasure, not to mention totally reasonable conversions. Why?

2996_reflection/reflection.md
```diff
@@ -3477,6 +3497,17 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
 
 ### [expr.const]{.sref} Constant Expressions {-}
 
+Modify (and clean up) the definition of _immediate-escalating_ to also apply to expressions of consteval-only type.
+
+::: std
+[18]{.pnum} A[n]{.rm} [potentially-evalauted]{.addu} expression or conversion is _immediate-escalating_ if it is [not]{.rm} [neither]{.addu} initially in an immediate function context [nor a subexpression of an immediate invocation,]{.addu} and it [is]{.rm} either
```
brevzin, line 3503: typo: evalauted

2996_reflection/reflection.md
```diff
@@ -5428,7 +5459,7 @@ consteval bool is_data_member_spec(info r);
   - each `$K$`^th^ reflection value in `mdescrs` describes a data member with all of the same properties as the `$K$`^th^ data member of `$C$`.
 - [#.#]{.pnum} `is_data_member_spec(@$r$~$K$~@)` is `true` for every `@$r$~$K$~@` in `mdescrs`,
 - [#.#]{.pnum} the type represented by `type_of(@$r$~$K$~@)` is a complete type for every `@$r$~$K$~@` in `mdescrs`, and
-- [#.#]{.pnum} for every pair 0 ≤ `$K$` < `$L$` < `mdescrs.size()`,  if `has_identifier(@$r$~$K$~@) && has_identifier(@$r$~$L$~@) && u8identifier_of(@$r$~$K$~@) != u8"_" && u8identifier_of(@$r$~$L$~@) != u8"_"` is `true`, then `u8identifier_of(@$r$~$K$~@) != u8identifier_of(@$r$~$L$~@)`. [Every provided identifier that is not `"_"` needs to be unique.]{.note}
+- [#.#]{.pnum} for every pair 0 ≤ `$K$` < `$L$` < `mdescrs.size()`,  if `has_identifier(@$r$~$K$~@) && has_identifier(@$r$~$L$~@)`, then either `u8identifier_of(@$r$~$K$~@) != u8identifier_of(@$r$~$L$~@)` or `u8identifier_of(@$r$~$K$~@) == u8"_"`. [Every provided identifier that is not `"_"` must be unique.]{.note}
```
brevzin, line 5462: Yeah that's shorter :-)

2996_reflection/reflection.md
brevzin: What's the problem we're trying to solve here? Okay, so I have a `void*` to an `info`. What am I going to do with it? 

```cpp
constexpr meta::info x = ^^int;
int main() {
    void* p = (void*)&x;
    // ???
}
```

All operations on this thing are `consteval` so it's not like I can actually call anything, since `p` isn't constant. What else can I do with the thing? `bit_cast` it to `u64` or something?

katzdm: Pretty much. I suppose we can say that the object representation of all `std::meta::info` values is unspecified, and maybe...that's good enough?

brevzin: So I'm currently rewriting P3032 to do immediate-escalating expressions and I think we want to say something like this (terms TBD)

> An expression is consteval-only if it is:
> 
> * an id-expression naming a consteval function,
> * an explicit or implicit call to a consteval function, or
> * has consteval-only type or is an explicit or implicit conversion to **or from** a consteval-only type.

Then something like:

> An expression is *immediate-escalating* if it is a consteval-only expression unless:
> 
> * it is initially in an immediate function context, or
> * it either is a constant expression or is a subexpression of a constant expression.

And then I think the rest of the words kind of work out? 

I'm actually not 100% sure on the "or from" part. But is it really feasible to have a conversion *from* a consteval-only type that *is not* a constant expression/

2996_reflection/reflection.md
```diff
@@ -3080,16 +3080,20 @@ Add a new paragraph at the end of [basic.types.general]{.sref} as follows:
...
-  - a pointer-to-member type to a class `C` of type `M` where either `C` or `M` is a consteval-only type.
+  - [#.#]{.pnum} `std::meta::info`, or
+  - [#.#]{.pnum} a pointer or reference to a consteval-only type,
+  - [#.#]{.pnum} an (possibly multi-dimensional) array of a consteval-only type,
+  - [#.#]{.pnum} a class type with a base class or non-static data member of consteval-only type,
+  - [#.#]{.pnum} a function type with a return type or parameter type of consteval-only type, or
+  - [#.#]{.pnum} a pointer-to-member type to a class `C` of type `M` where either `C` or `M` is a consteval-only type.
 
-An object of consteval-only type shall either end its lifetime during the evaluation of a manifestly constant-evaluated expression or conversion ([expr.const]{.sref}), or be a constexpr variable for which every expression that names the variable is within an immediate function context.
+[13]{.pnum} Every object of consteval-only type shall either be
```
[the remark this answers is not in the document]
katzdm: Done.

2996_reflection/reflection.md
katzdm: Addressed this with what we reached on Monday's call: consteval-only-erased pointers and references aren't permitted results of constant expressions.

2996_reflection/reflection.md
```diff
@@ -3239,21 +3309,27 @@ Add a new subsection of [expr.prim]{.sref} following [expr.prim.req]{.sref}
...
+[#]{.pnum} The `$splice-specifier$` designates either
+
+* [#.#]{.pnum} a value or object,
+* [#.#]{.pnum} a variable,
+* [#.#]{.pnum} a structured binding,
+* [#.#]{.pnum} a function,
+* [#.#]{.pnum} an enumerator, or
+* [#.#]{.pnum} a non-static data member.
+
+The `$splice-specifier$` shall not designate an unnamed bit-field, a constructor, or a destructor.
```
brevzin, line 3326: I think this should just go into "a function other than a constructor or destructor" and then unnamed bit-field can be a note after the last bullet that it's not a non-static data member.

2996_reflection/reflection.md
```diff
@@ -3271,41 +3347,43 @@ Add a production to `$postfix-expression$` for splices in member access expressi
...
+Modify paragraph 1 to account for splices in member access expressions. Prefer the word "possibly" to "optionally" since `template` cannot precede a `splice-expression`:
 
 ::: std
-[1]{.pnum} A postfix expression followed by a dot `.` or an arrow `->`, optionally followed by the keyword template, and then followed by an _id-expression_[, or a _splice-expression_ designating a class member]{.addu}, is a postfix expression. [If the keyword `template` is used, the following unqualified name is considered to refer to a template ([temp.names]). If a `$simple-template-id$` results and is followed by a `​::`​, the _id-expression_ [or _splice-expression_]{.addu} is a qualified-id.]{.note}
+[1]{.pnum} A postfix expression followed by a dot `.` or an arrow `->`, [optionally]{.rm} [possibly]{.addu} followed by the keyword template, and then followed by an `$id-expression$` [or a splicer designating a class member]{.addu}, is a postfix expression. [If the keyword `template` is used, the following unqualified name is considered to refer to a template ([temp.names]). If a `$simple-template-id$` results and is followed by a `::`, the `$id-expression$` is a `$qualified-id$`.]{.note}
```
brevzin, line 3363: code font for `template` (preexisting)

2996_reflection/reflection.md
```diff
@@ -3365,21 +3443,18 @@ Add a new subsection of [expr.unary]{.sref} following [expr.delete]{.sref}
...
 $reflect-expression$:
    ^ ::
-   ^ $nested-name-specifier$@~_opt_~@ $namespace-name$
-   ^ $nested-name-specifier$@~_opt_~@ $template-name$
-   ^ $nested-name-specifier$@~_opt_~@ $concept-name$
+   ^ $unqualified-id$
+   ^ $qualified-id$
    ^ $type-id$
-   ^ $id-expression$
+   ^ $pack-index-expression$
```
brevzin, line 3452: Okay so before we had id-expression and then all the names separately. 

Now you removed the names and id-expression, and replaced it with... all the constituents of id-expression (unqualified-id, qualified-id, and pack-index-expression). Can we go back to id-expression then?

Why did we drop the namespace/template/concept name stuff? I don't think those are (un)qualified-ids.

2996_reflection/reflection.md
```diff
@@ -3477,49 +3547,67 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
-Add new paragraphs prior to the definition of _manifestly constant evaluated_ ([expr.const]{.sref}/20), and renumber accordingly:
+Modify paragraph 15 to disallow returning non-consteval-only pointers and references to consteval-only objects from constant expressions.
 
 ::: std
-::: addu
+[15]{.pnum} A _constant-expression_ is either a glvalue core constant expression that [\ ]{.addu}
 
-[20]{.pnum} An expression or conversion is _in substitution context_ if it results from the substitution of template parameters
+* [#.#]{.pnum} refers to an entity that is a permitted result of a constant expression[, and]{.addu}
+* [[#.#]{.pnum} is of consteval-only type if the entity designated by the expression is of consteval-only type,]{.addu}
```
brevzin, line 3556: Add a note that this prohibits forming a `Base&` to a `Derived` if the latter is consteval-only but the former is not.

2996_reflection/reflection.md
```diff
@@ -3477,49 +3547,67 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
 
-* [#.#]{.pnum} during template argument deduction ([temp.deduct]{.sref}),
-* [#.#]{.pnum} in a `$concept-id$` ([temp.names]{.sref}), or
-* [#.#]{.pnum} in a `$requires-expression$` ([expr.prim.req]{.sref}).
+or a prvalue core constant expression whose value satisfies the following constraints:
+
+* [#.#]{.pnum} if the value is an object of class type, each non-static data member of reference type refers to an entity that is a permitted result of a constant expression [and has consteval-only type if the entity has consteval-only type]{.addu},
```
brevzin, line 3560:
```suggestion
* [#.#]{.pnum} if the value is an object of class type, each non-static data member of reference type refers to an entity that is a permitted result of a constant expression [and has consteval-only type if that entity has consteval-only type]{.addu},
* ```

File: 2996_reflection/reflection.md [hunk 11]
```diff
@@ -3477,49 +3547,67 @@ Add a new paragraph between [expr.eq]{.sref}/5 and /6:
...
+* [#.#]{.pnum} if the value is an object of scalar type, it does not have an indeterminate or erroneous value ([basic.indet]),
+* [#.#]{.pnum} if the value is of pointer type, it is a pointer to an object of static storage duration, a pointer past the end of such an object ([expr.add]), a pointer to a non-immediate function, or a null pointer value,
+* [[#.#]{.pnum} if the value is a pointer to an object of consteval-only type, or a pointer past the end of such an object, then the prvalue is also of consteval-only type,]{.addu}
```
brevzin, line 3563: this should be part of the previous bullet somehow. Maybe like:

> * if the value is of pointer type, then:
>   * it is a pointer to, or past the end of, object of static storage duration and, if the object has consteval-only type, the value also has consteval-only type, or
>   * it is a pointer to a non-immediate function, or
>   * it is a null pointer value,

Ish?

2996_reflection/reflection.md
brevzin: Add a `[...]{.ednote}` explaining this change.

2996_reflection/reflection.md
```diff
@@ -3189,21 +3208,18 @@ Add a production to the grammar for `$nested-name-specifier$` as follows:
...
 ```
 :::
 
-Add a new paragraph restricting `$splice-namespace-qualifier$`, and renumber accordingly:
+Add a new paragraph restricting `$splice-specifier$`, and renumber accordingly:
 
 ::: std
 ::: addu
-[0]{.pnum} The `$splice-specifier$` of a `$splice-namespace-qualifier$` shall designate a namespace or namespace alias.
+[0]{.pnum} A `$splice-specifier$` of a `$nested-name-specifier$` either is dependent or designates a type, `$typedef-name$`, namespace, or `$namespace-alias$`.
```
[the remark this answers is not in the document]
katzdm: Re: shall, "The" - sgtm, fixed.

Re: "even if it's dependent, it still needs to designate the thing" - I think, possibly not? But I drew inspiration from wording in [[basic.lookup.qual](https://eel.is/c++draft/basic.lookup.qual#general-1.sentence-2)]:

> If a name, template-id, or computed-type-specifier is followed by a ​::​, it shall designate a namespace, **class**, enumeration, **or dependent type**, and the ​::​ is never interpreted as a complete nested-name-specifier.

2996_reflection/reflection.md
```diff
@@ -3143,40 +3170,86 @@ Change the grammar for `$primary-expression$` in [expr.prim]{.sref} as follows:
...
+
+In all other cases, it is interpreted as a `$splice-expr-template-specifier$`.
+
+[#]{.pnum} The `typename` in a `$splice-type-specifier$` may only be omitted in a type-only context, or when the `$splice-specifier$` is preceded by `template`.
+
+::: example
+```cpp
+int var {};
+struct cls {};
+template<int> int t_var {};
```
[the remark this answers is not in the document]
katzdm: Done.

2996_reflection/reflection.md
```diff
@@ -3143,40 +3170,86 @@ Change the grammar for `$primary-expression$` in [expr.prim]{.sref} as follows:
...
+  { typename [:TCls:]<0> *var; }           // ok: declaration of 't_cls<0> *var'.
+  { template typename [:TCls:]<0> *var; }  // also ok.
+
+  { template [:Concept:] auto *var = 0; }  // ok: deduced placeholder type.
+};
+
+void non_dependent() {
+  dependent<^var, ^cls, ^t_var, ^t_cls, ^always_true>();
+
+  { template [:^^t_cls:]<0> *var; }        // ok in non-dependent context.
```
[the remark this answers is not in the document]
katzdm: Done.

2996_reflection/reflection.md
```diff
@@ -3143,40 +3170,86 @@ Change the grammar for `$primary-expression$` in [expr.prim]{.sref} as follows:
...
+void dependent() {
+  { [:Var:] *var; }                        // ok: multiplication of 'var * var'.
+
+  { typename [:Cls:] *var; }               // ok: declaration of 'cls *var'
+  { template [:TVar:]<0> *var; }           // ok: multiplication of 't_var<0> * var'.
+  { template [:TCls:]<0> *var; }           // error: cannot splice type as expression.
+  { typename [:TCls:]<0> *var; }           // ok: declaration of 't_cls<0> *var'.
+  { template typename [:TCls:]<0> *var; }  // also ok.
+
+  { template [:Concept:] auto *var = 0; }  // ok: deduced placeholder type.
```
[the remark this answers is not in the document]
katzdm: Done - left the ones as `var` that I expect to be multiplication, though (since I want them to name an existing variable).

2996_reflection/reflection.md
katzdm: I like that, done.

2996_reflection/reflection.md
katzdm: Good catch, done.

2996_reflection/reflection.md
katzdm: Done.

> I don't think those are (un)qualified-ids

Hmm. We'll have to see what Core says here.

2996_reflection/reflection.md
katzdm: Good idea, done.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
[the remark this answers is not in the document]
brevzin: I realize here what I meant to do was just remove the extra `^` but also I changed all the spacing. Thanks, UI!

2996_reflection/reflection.md
```diff
@@ -3696,66 +3840,60 @@ Extend the grammar for `$using-enum-declarator$` as follows:
...
 ```
 :::
 
-Modify paragraph 1 of [enum.udecl]{.sref} as follows:
+Modify paragraph 1 to handle splicers:
 
 ::: std
 
-[1]{.pnum} A `$using-enum-declarator$` [that is not a `$splice-specifier$`]{.addu} names the set of declarations found by lookup ([basic.lookup.unqual]{.sref}, [basic.lookup.qual]{.sref}) for the `$using-enum-declarator$`. The `$using-enum-declarator$` shall designate a non-dependent type with a reachable `$enum-specifier$`.
+[1]{.pnum} [A `using-enum-declarator` of the form `$splice-type-specifier$` is considered in a type-only context, and designates the same construct designated by the splicer. Any other]{.addu} [A]{.rm} `$using-enum-declarator$` names the set of declarations found by type-only lookup ([basic.lookup.general]) for the `$using-enum-declarator$` ([basic.lookup.unqual], [basic.lookup.qual]). The `$using-enum-declarator$` shall designate a non-dependent type with a reachable `$enum-specifier$`.
```
brevzin, line 3851: Missing $s around `using-enum-declarator`. Also "is a type-only context" rather than "is considered in a".

Also, shouldn't that part go into [temp] where we define what type-only context is? I think it'd be nice to have all those localized.

2996_reflection/reflection.md
```diff
@@ -3696,66 +3840,60 @@ Extend the grammar for `$using-enum-declarator$` as follows:
...
++     $attribute-specifier-seq$@~_opt_~@ using namespace $splice-specifier$
 ```
 :::
 
 Add the following prior to the first paragraph of [namespace.udir]{.sref}, and renumber accordingly:
 
 ::: std
 ::: addu
-[0]{.pnum} The `$qualified-namespace-specifier$` shall neither contain a dependent `$nested-name-specifier$` nor a dependent `$splice-specifier$`.
+[0]{.pnum} The `$splice-specifier$`, if any, designates a namespace or `$namespace-alias$`. Neither the `$nested-name-specifier$` nor the `$splice-specifier$` shall be dependent.
```
brevzin, line 3896: I think flip it - The nested-name-specifier and the splice-specifier shall not be dependent.

2996_reflection/reflection.md
```diff
@@ -3820,14 +3971,58 @@ Modify the definition of reachability to account for injected declarations:
...
+::: std
+...
+
+A class declaration where the `$class-name$` in the `$class-head-name$` is a `$simple-template-id$` shall be an explicit specialization ([temp.expl.spec]) or a partial specialization ([temp.spec.partial]). [The `$simple-template-id$` shall not contain a splicer.]{.addu} A `$class-specifier$` whose `$class-head$` omits the `$class-head-name$` defines an _unnamed class_.
+
+:::
+
+### [class.name] Class names {-}
+
+Cover `$splice-type-specifier$s in paragraph 5.
```
brevzin, line 3987: Missing close tick

2996_reflection/reflection.md
```diff
@@ -3944,8 +4164,8 @@ Adjust paragraph 3 of [temp.arg.general] to not apply to splice template argumen
     f<int()>();       // int() is a type-id: call the first f()
 
 +   constexpr int x = 42;
-+   f<[:^int:]>();    // splice-template-argument: calls the first f()
-+   f<[:^x:]>();      // splice-template-argument: calls the second f()
++   f<[:^int:]>();    // splice-specifier: calls the first f()
++   f<[:^x:]>();      // splice-specifier: calls the second f()
```
brevzin, line 4168: Can we just change this example to add `#1` and `#2` to the two fs so that we can say calls `#1`, calls `#2`? I think we usually do that.

2996_reflection/reflection.md
```diff
@@ -3143,40 +3170,86 @@ Change the grammar for `$primary-expression$` in [expr.prim]{.sref} as follows:
...
+```cpp
+int var {};
+struct cls {};
+template<int> int t_var {};
+template<int> struct t_cls { };
+template <typename> concept always_true = requires { true; };
+
+template <auto Var, auto Cls, auto TVar, auto TCls, auto Concept>
+void dependent() {
+  { [:Var:] *var; }                        // ok: multiplication of 'var * var'.
```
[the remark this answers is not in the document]

2996_reflection/reflection.md
katzdm: It's kinda weird that `using-enum-declarator` _isn't_ listed in the `[temp.res.general]` list already? I'd almost rather leave it as-is, and learn during CWG review if there's a reason why.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Good catch, thanks.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
```diff
@@ -3212,21 +3212,21 @@ template <typename> concept always_true = requires { true; };
...
 void dependent() {
-  [:Var:] *var;                      // ok: multiplication of 'var * var'.
+  [:Var:] *var;                      // OK, multiplication of 'var * var'
 
-  typename [:Cls:] *a;               // ok: declaration of 'cls *var'
-  template [:TVar:]<0> *var;         // ok: multiplication of 't_var<0> * var'.
-  template [:TCls:]<0> *var;         // error: cannot splice type as expression.
-  typename [:TCls:]<0> *b;           // ok: declaration of 't_cls<0> *var'.
-  template typename [:TCls:]<0> *c;  // also ok.
+  typename [:Cls:] *a;               // OK, declaration of 'cls *var'
```
brevzin, line 3217:
```suggestion
  typename [:Cls:] *a;               // OK, declaration of 'cls *a'
```

2996_reflection/reflection.md
```diff
@@ -3212,21 +3212,21 @@ template <typename> concept always_true = requires { true; };
...
+  template [:TVar:]<0> *var;         // OK, multiplication of 't_var<0> * var'
+  template [:TCls:]<0> *var;         // error: cannot splice type as expression
+  typename [:TCls:]<0> *b;           // OK, declaration of 't_cls<0> *var'
```
brevzin, line 3220:
```suggestion
  typename [:TCls:]<0> *b;           // OK, declaration of 't_cls<0> *b'
```

2996_reflection/reflection.md
```diff
@@ -4157,15 +4157,15 @@ Adjust paragraph 3 of [temp.arg.general] to not apply to splice template argumen
 
 ::: example2
 ```diff
-  template<class T> void f();
-  template<int I> void f();
+  template<class T> void f(); @[\ \ // #1]{.addu}@
+  template<int I> void f(); @[\ \ \  // #2]{.addu}@
```
brevzin, line 4161: Heh, I would just done like
```diff
- template<class T> void f();
- template<int T> void f();
+ template<class T> void f() // #1
+ template<int T> void f()   // #2
```
But that works too.

2996_reflection/reflection.md
katzdm: Done.

PR #208, opened 2024-12-11
Audit id-expression. Fix "struct to struct of arrays" example.

katzdm:
Mainly two things.
1. I audited every use of _id-expression_ in the Standard, and found a bunch of places where we need to account for _splice-expressions_.
2. The "struct to struct of arrays" example is broken in the face of the new injected declaration rules (i.e., it injected a declaration from an instantiation into the global namespace) - this "fixes" the example against our latest understanding of those rules.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -846,25 +846,25 @@ On Compiler Explorer: [EDG](https://godbolt.org/z/W74qxqnhf), [Clang](https://go
 #include <meta>
 #include <array>
 
-template <typename T, std::size_t N>
-struct struct_of_arrays_impl;
+template <typename T, size_t N>
+constexpr auto struct_of_arrays_impl = [] consteval {
+  struct impl;
 
-consteval auto make_struct_of_arrays(std::meta::info type,
-                                     std::meta::info N) -> std::meta::info {
-  std::vector<std::meta::info> old_members = nonstatic_data_members_of(type);
+  std::vector<std::meta::info> old_members = nonstatic_data_members_of(^^T);
   std::vector<std::meta::info> new_members = {};
   for (std::meta::info member : old_members) {
-    auto type_array = substitute(^^std::array, {type_of(member), N });
-    auto mem_descr = data_member_spec(type_array, {.name = identifier_of(member)});
+    auto array_type = substitute(^^std::array, {
+        type_of(member),
+        std::meta::reflect_value(N),
+    });
+    auto mem_descr = data_member_spec(array_type, {.name = identifier_of(member)});
     new_members.push_back(mem_descr);
   }
-  return define_aggregate(
-    substitute(^^struct_of_arrays_impl, {type, N}),
-    new_members);
-}
+  return std::meta::define_aggregate(^^impl, new_members);
+}();
 
 template <typename T, size_t N>
-using struct_of_arrays = [: make_struct_of_arrays(^^T, ^^N) :];
+using struct_of_arrays = [: struct_of_arrays_impl<T, N> :];
 ```
 :::
 
@@ -890,7 +890,7 @@ using points = struct_of_arrays<point, 30>;
 
 Again, the combination of `nonstatic_data_members_of` and `define_aggregate` is put to good use.
 
-On Compiler Explorer: [EDG](https://godbolt.org/z/5nsqY1vxz), [Clang](https://godbolt.org/z/veEq6ozes).
+On Compiler Explorer: [EDG](https://godbolt.org/z/jWrPGhn5s), [Clang](https://godbolt.org/z/7PcYrY7eq).
 
 
 ## Parsing Command-Line Options II
@@ -2921,6 +2921,17 @@ Education and training are important to help C++ users avoid such sharp edges, b
 
 ## Language
 
+### [intro.defs]{.sref} Terms and definitions {-}
+
+Add `$splice-specifier$` to the list of template argument forms in definition 3.5.
+
+::: std
+**[3.5]{.pnum} argument**
+
+⟨template instantiation⟩ `$constant-expression$`, `$type-id$`, [or]{.rm} `$id-expression$`[, or `$splice-specifier$`]{.addu} in the comma-separated list bounded by the angle brackets
+
+:::
+
 ### [lex.phases]{.sref} Phases of translation {-}
 
 [In addition to changes necessary for this proposal, we are applying the "drive-by fix" of merging phases 7/8, in order to clarify that template instantiation is interleaved with translation. In so doing, we replace the notion of "instantiation units" with a partial ordering among all program constructs in a translation unit.]{.ednote}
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -846,25 +846,25 @@ On Compiler Explorer: [EDG](https://godbolt.org/z/W74qxqnhf), [Clang](https://go
 #include <meta>
 #include <array>
 
-template <typename T, std::size_t N>
-struct struct_of_arrays_impl;
+template <typename T, size_t N>
+constexpr auto struct_of_arrays_impl = [] consteval {
```
brevzin, line 850: I don't think we should need this `consteval` right? Doesn't hurt, but shouldn't be strictly necessary?

2996_reflection/reflection.md
```diff
@@ -846,25 +846,25 @@ On Compiler Explorer: [EDG](https://godbolt.org/z/W74qxqnhf), [Clang](https://go
...
-  return define_aggregate(
-    substitute(^^struct_of_arrays_impl, {type, N}),
-    new_members);
-}
+  return std::meta::define_aggregate(^^impl, new_members);
+}();
 
 template <typename T, size_t N>
-using struct_of_arrays = [: make_struct_of_arrays(^^T, ^^N) :];
+using struct_of_arrays = [: struct_of_arrays_impl<T, N> :];
```
brevzin, line 867: I think we absolutely need an example or two illustrating what about the original is broken. Also, doesn't this have consequences for linkage? Originally, `struct_of_arrays<T, N>` had external linkage, now doesn't it have... no linkage?

2996_reflection/reflection.md
```diff
@@ -3067,6 +3078,29 @@ Since namespace aliases are now entities, but their declarations are not definit
 
 ### [basic.def.odr]{.sref} One-definition rule {-}
 
+Add `$splice-expression$`s to the set of potential results of an expression in paragarph 3.
```
brevzin, line 3081:
```suggestion
Add `$splice-expression$`s to the set of potential results of an expression in paragraph 3.
```

2996_reflection/reflection.md
```diff
@@ -3067,6 +3078,29 @@ Since namespace aliases are now entities, but their declarations are not definit
...
+
+::: std
+[3]{.pnum} An expression or conversion is _potentially evaluated_ unless it is an unevaluated operand ([expr.context]{.sref}), a subexpression thereof, or a conversion in an initialization or conversion sequence in such a context. The set of _potential results_ of an expression `$E$` is defined as follows:
```
brevzin, line 3084: sref!

2996_reflection/reflection.md
```diff
@@ -3067,6 +3078,29 @@ Since namespace aliases are now entities, but their declarations are not definit
...
+- [#.#]{.pnum} [...]
+
+[This set is a (possibly-empty) set of `$id-expression$`s [and `$splice-expression$`s]{.addu}, each of which is either `$E$` or a subexpression of `$E$`.]{.note}
+
+:::
+
+Break bullet 4.1 into sub-bullets, modify it to cover splicing of functions, and replace [basic.lookup] with [over.pre] since the canonical definition of "overload set" is relocated there by this proposal:
+
+::: std
+- [4.1]{.pnum} A function is named by an expression or conversion if it is the selected member of an overload set ([[basic.pre]]{.rm} [[over.pre]]{.addu}, [over.match], [over.over]) in an overload resolution performed as part of forming that expression or conversion, unless it is a pure virtual function and [either]{.rm} the expression[\ ]{.addu}
```
brevzin, line 3096: I don't actually see `[basic.pre]`? Right now it's `[basic.lookup]`. 

Also what's with the backslash?

2996_reflection/reflection.md
```diff
@@ -3067,6 +3078,29 @@ Since namespace aliases are now entities, but their declarations are not definit
...
+
+  - [#.#.#]{.pnum} is not an `$id-expression$` naming the function with an explicitly qualified name[,]{.addu}
+  - [[#.#.#]{.pnum} is not a `$splice-expression$`,]{.addu} or
+  - [#.#.#]{.pnum} [the expression]{.rm} forms a pointer to member ([expr.unary.op]).
```
brevzin, line 3100: I think move "the expression" into each bullet. Even better, "A function is named by an expression or conversion [`$E$`]{.addu} if [...]" and then:

> * [#.#.#]{.pnum} [`$E$`]{.addu} is not an `$id-expression$` [...],
> * [#.#.#]{.pnum} [`$E$` is not a `$splice-expression$`, or]{.addu}
> * [#.#.#]{.pnum} [the expression]{.rm} [`$E$`]{.addu} forms a pointer to member.

?

2996_reflection/reflection.md
```diff
@@ -3436,6 +3489,18 @@ the `$id-expression$` is transformed into a class member access expression using
...
+
+::: std
+[4]{.pnum} An _implicitly movable entity_ is a variable of automatic storage duration that is either a non-volatile object or an rvalue reference to a non-volatile object type. In the following contexts, [an]{.rm} [a (possibly parenthesized)]{.addu} `$id-expression$` [or `$splice-expression$` ([expr.prim.splice]) `$E$`]{.addu} is _move-eligible_:
+
+- [#.#]{.pnum} If [the `$id-expression$` (possibly parenthesized)]{.rm} [`$E$`]{.addu} is an operand of a `return` ([stmt.return]) or `co_return` ([stmt.return.coroutine]) statement, and names an implicitly movable entity declared in the body or `$parameter-declaration-clause$` of the innermost enclosing function or `$lambda-expression$`, or
+- [#.#]{.pnum} if [the `$id-expression$` (possibly parenthesized)]{.rm} [`$E$`]{.addu} is the operand of a `$throw-expression$` ([expr.throw]) and names an implicitly movable entity that belongs to a scope that does not contain the `$compound-statement$` of the innermost `$lambda-expression$`, `$try-block$`, or `$function-try-block$` (if any) whose `$compound-statement$` or `$ctor-initializer$` contains the `$throw-expression$`.
```
brevzin, line 3500: 👍 So basically I'm suggesting that you do the same thing in the other section that you already did here.

2996_reflection/reflection.md
```diff
@@ -3513,6 +3578,23 @@ Break the next paragraph into a bulleted list, extend it to also cover splices,
 
 :::
 
+### [expr.prim.lambda.capture]{.sref} Captures {-}
+
+Modify bullet 7.1 as follows:
+
+::: std
+- [7.1]{.pnum} An `$id-expression$` [or `$splice-expression$`]{.addu} that names a local entity potentially references that entity; an `$id-expression$` that names one or more non-static class members and does not form a pointer to member ([expr.unary.op]) potentially references `*this`.
```
brevzin, line 3586: Should the second `$id-expression$` also get the "or `$splice-expression$`" treatment? If not, why not?

```cpp
struct C {
  int i;

  void f() {
    constexpr auto r = ^^i;
    [this]{
      f([:r:]); // is this... 
      f(i); // ... equivalent to this?
    };
  }
};
```

2996_reflection/reflection.md
```diff
@@ -3747,7 +3829,7 @@ is interpreted as follows:
 
   * [#.#]{.pnum} If the `$id-expression$` denotes an overload set `$S$` and overload resolution for the expression `&$S$` determines a unique function `$F$` ([over.over]{.sref}), the `$reflect-expression$` represents `$F$`.
 
-  * [#.#]{.pnum} Otherwise, if the `$id-expression$` denotes a variable, structured binding, enumerator, or non-static data member or member function, the `$reflect-expression$` represents that entity.
+  * [#.#]{.pnum} Otherwise, if the `$id-expression$` denotes a variable or structured binding not captured by an enclosing `$lambda-expression$`, an enumerator, or a non-static data member or member function, the `$reflect-expression$` represents that entity.
```
brevzin, line 3832: Would prefer splitting out the capturing part.

2996_reflection/reflection.md
```diff
@@ -4030,6 +4119,24 @@ Add a row to [tab:dcl.type.simple] to cover the `$splice-type-specifier$` produc
...
+
+Add a bullet after bullet 1.3 to apply to `$splice-expression$`s:
+
+::: std
+[1]{.pnum} For an expression `$E$`, the type denoted by `decltype($E$)` is defined as follows:
+
+[...]
+
+- [1.3]{.pnum} otherwise, if `$E$` is an unparenthesized `$id-expression$` or an unparenthesized class member access ([expr.ref]), `decltype($E$)` is the type of the entity named by `$E$`. If there is no such entity, the program is ill-formed;
+- [[1.3+]{.pnum} otherwise, if `$E$` is an unparenthesized `$splice-expression$`, `decltype($E$)` is the type of the entity, object, or value named by `$E$`;]{.addu}
```
brevzin, line 4132: Is the only reason this isn't folded into the previous bullet that there it's "entity named by E" and here it's "entity, object, or value named by E"?

2996_reflection/reflection.md
katzdm: Sorry, the godbolt links have a better version. Replaced, done.

2996_reflection/reflection.md
katzdm: ^ replaced with better version from the godbolt links that shouldn't have the linkage issue.

I have a list of examples to add; I'll add that to it.

2996_reflection/reflection.md
katzdm: Done

2996_reflection/reflection.md
katzdm: Lmaooo I tried to be good, but I slipped up. Done.

2996_reflection/reflection.md
katzdm: Good catch, thanks.

In a few places where I've broken a clause into a bulleted list, I've put an explicitly highlighted space at the end of the line to call out that the line break wasn't there before.

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Donezo

2996_reflection/reflection.md
katzdm: No, because _splice-expression_ doesn't undergo the implicit `(*this)`-transformation; you must write `this->[:r:]`.

2996_reflection/reflection.md
katzdm: Right, thanks for the reminder. Done.

2996_reflection/reflection.md
katzdm: Ah, not quite - it should be "designated the _splice-specifier_ of" not "named by". Thanks.

brevzin (approved):

PR #209, opened 2024-12-12
Always be updating

katzdm:
- Also IFNDR for "indeterminately sequenced" evaluations
- Polish examples of rules for injected declarations given in [expr.const]
- Informal section discussing restrictions on injected declarations

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -4005,7 +4005,7 @@ Modify (and clean up) the definition of _immediate-escalating_ in paragraph 18 t
 
 :::
 
-Add a new paragraphs prior to the definition of _manifestly constant evaluated_ ([expr.const]{.sref}/21), and renumber accordingly:
+Add a new paragraphs prior to the definition of _manifestly constant-evaluated_ ([expr.const]{.sref}/21), and renumber accordingly:
 
 ::: std
 ::: addu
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -2283,6 +2284,86 @@ Lastly, we clarify that during the definition of an _injected declaration_, the
...
+  }
+};
+
+static_assert(TCls<void>::sfn());
+static_assert(is_complete_type(^^S));
+```
+
+The above example observes the effects of the failed substitution of `#1` by way of the completeness of `S`. Such tricks can be used to observe implementation details, like the order in which overloads are checked, that may be unportable (and which implementations might desire to change over time).
+
+Our proposed solution, specified in [expr.const]/23.2, is to make it ill-formed to produce an injected declaration _from_ a manifestly constant-evaluated expression within an instantiation to _outside of_ that instantiation, or visa versa. Because that expression in the example above (`define_aggregate(^^S, {})`) is within the instantiation of the requires clause of `TCls<void>::sfn`, and the target scope of the injected declaration is outside of that same instantiaton, the example becomes ill-formed (diagnostic required). Note that this does not prevent writing `consteval` function templates that wrap `define_aggregate`:
```
brevzin, line 2316: I think better to explicitly say: "produce an injected declaration _from inside of_ a manifestly constant-evaluated expression within an instantiation to _outside of_ that instantiation". Specifically having both INSIDE and OUTSIDE italicized helps make that clear.

2996_reflection/reflection.md
```diff
@@ -2283,6 +2284,86 @@ Lastly, we clarify that during the definition of an _injected declaration_, the
...
+    return true;   // always selected
+  }
+};
+
+static_assert(TCls<void>::sfn());
+```
+
+Athough the instantiation of `TCls1<void>` in the requires-clause of `#1` causes an injected declaration to be produced, it is not discernibly a side-effect of the failed substitution: Observing the side effect will first require one to write (some moral  equivalent of) `TCLs1<void>::Incomplete`, the act of which would otherwise itself trigger the same side-effect.
+
+Although this rule constrains the manner with which `define_aggregate` can be used, we are not aware of any motivating use cases for P2996 that are harmed. Worth mentioning, however: the rule has more dire implications for other code injection papers being considered by WG21, most notably P3294 ("_Code Injection With Token Sequences_"). With this rule as it is, it becomes impossible for e.g., the instantiation of a class template specialization `TCls<Foo>` to produce an injected declaration of `std::formatter<TCls<Foo>>` (since the target scope would be the global namespace).
```
brevzin, line 2361: Use [@P3294R2] instead of just P3294, which will add the link.

2996_reflection/reflection.md
```diff
@@ -2283,6 +2284,86 @@ Lastly, we clarify that during the definition of an _injected declaration_, the
...
+
+In this context, we do believe that relaxations of the rule can be considered: For instance, we ought to be able to say that the instantiation of `std::formatter<TCls<Foo>>` is sequenced strictly after the instantiation of `TCls<Foo>`, and observations such as these might make it possible to permit such injections without making it "discernible" whether they resulted from failed substitutions. The key to such an approach would be to define a partial order over the instantiations of a program, and to extend the _semantically sequenced_ relation introduced by P2996 ([lex.phases]/7) to apply to constructs _across_ instantiations when the relative order of their respective instantiations is defined.
```
brevzin, line 2363: say "this paper" or "this proposal" instead of P2996.

2996_reflection/reflection.md
```diff
@@ -2283,6 +2284,86 @@ Lastly, we clarify that during the definition of an _injected declaration_, the
...
+
+All of that said, these relaxations are not needed for the code injection introduced by P2996, and we do not seek to introduce them at this time.
```
brevzin, line 2365: ditto not just P2996

2996_reflection/reflection.md
katzdm: Done.

2996_reflection/reflection.md
katzdm: Yep, thanks.

2996_reflection/reflection.md
katzdm: Done

2996_reflection/reflection.md
katzdm: Done

brevzin (approved):

PR #238, opened 2025-02-10
CWG 2025-02-10 post-meeting fixits.

katzdm:


2996_reflection/reflection.md
```diff
@@ -5734,6 +5736,38 @@ int a = fn<^^NS>();
...
+[1]{.pnum} A concept is dependent if it is designated by a dependent `$splice-specifier$`. A `$concept-name$` is dependent if its lookup context is dependent.
+
+[2]{.pnum} The program is ill-formed if any construct names or designates a dependent concept.
+
+::: example
+```cpp
+template <std::meta::info C, std::meta::info N>
+struct S {
+  void f() requires template [:C:]<S> {}       // error: concept designated by dependent splice
+  void g() requires [:N:]::template C<S> {}    // error: dependent concept-name
```
brevzin, line 5756: Do you think it's worth adding an example like...
```cpp
bool h() {
    return extract<bool>(substitute(C, {^^int})); // ok, returns true
}
```
Help illustrate that the issue is dependent splicing, not literally any dependent use at all.

2996_reflection/reflection.md
```diff
@@ -5734,6 +5736,38 @@ int a = fn<^^NS>();
...
 
+### 13.8.3.7 [temp.dep.concept] Dependent concepts {-}
+
+Add a new section to cover dependent concepts.
+
+::: std
+:::addu
+**Dependent concepts   [temp.dep.concept]**
+
+[1]{.pnum} A concept is dependent if it is designated by a dependent `$splice-specifier$`. A `$concept-name$` is dependent if its lookup context is dependent.
```
brevzin, line 5747: I think in light of P2841 being in flight we might want to make the term more specific here. Like maybe say a concept is *splice-dependent* if that, and a program is ill-formed we name/designate a splice-dependent concept?

2996_reflection/reflection.md
```diff
@@ -6978,9 +7012,9 @@ template <class T>
 
 [#]{.pnum} *Constant When*:
 
-- [#.#]{.pnum} If `r` represents a non-static data member of a class `C` with type `X`, then when `T` is `X C::*` and `r` does not represent a bit-field.
-- [#.#]{.pnum} Otherwise, if `r` represents an implicit object member function of class `C` with type `F` or `F noexcept`, then when `T` is `F C::*`.
-- [#.#]{.pnum} Otherwise, `r` represents a function, static member function, or explicit object member function of function type `F` or `F noexcept`, then when `T` is `F*`.
+- [#.#]{.pnum} `r` represents a non-static data member of a class `C` with type `X`, and `T` is `X C::*` and `r` does not represent a bit-field,
```
brevzin, line 7015: Drop the comma and I think replace with a semicolon?
```suggestion
- [#.#]{.pnum} `r` represents a non-static data member of a class `C` with type `X` and `T` is `X C::*` and `r` does not represent a bit-field;
```
Ditto the other two.

PR #254, opened 2025-04-17
[P2996] Fix wrong reference URL link to P3554R0

nofe1248:


brevzin:
Thanks! These are no longer necessary now that they're in the mailing, so I just removed them entirely in https://github.com/brevzin/cpp_proposals/commit/8ffbc56f4b45c38549f4ac7d274b6898164bb009.

PR #256, opened 2025-04-22
Add Tomasz's example to [temp.dep.splice].

katzdm:


2996_reflection/reflection.md
```diff
@@ -5897,6 +5897,25 @@ int main() {
...
+```cpp
+template<template<class> class X>
+struct S {
+  typename [: ^^X :]<int, float> m; 
+};
+
+template<class> struct V1 {};
+template<class, class = int> struct V2 {};
+
+S<V1> s1; // error: type of S<V1>::m is invalid
```
brevzin, line 5913: Should this say something like `V1<int, float>` is an invalid template-id, or something more specific like that?

katzdm: Went with "too many template arguments", since it's technically a _splice-specialization-specifier_.

brevzin: Works for me.

PR #260, opened 2025-05-01
Post-CWG updates for 2025-04-25.

katzdm:
Further edits to `access_context::current()` and `is_accessible`.

Other minor edits:
- Move `type_of` and friends earlier in `[meta.reflection]` such that the definition of _has-type_ precedes its usage in e.g., `is_const`.
- "same as that of _S_" instead of "same as _S_" (which makes it sound like _S_ is a type).
- Remove some more `{.sref}`s.

### DIFF
--- a/2996_reflection/reflection.md
+++ b/2996_reflection/reflection.md
@@ -4190,7 +4190,7 @@ auto g = typename [:^^int:](42);
 
   [The type of a `$splice-expression$` designating a variable or structured binding of reference type will be adjusted to a non-reference type ([expr.type]{.sref}).]{.note}
 
-* [#.#]{.pnum} Otherwise, if `$S$` is a value or an enumerator, the expression is a prvalue that computes `$S$` and whose type is the same as `$S$`.
+* [#.#]{.pnum} Otherwise, if `$S$` is a value or an enumerator, the expression is a prvalue that computes `$S$` and whose type is the same as that of `$S$`.
 
 * [#.#]{.pnum} Otherwise, the expression is ill-formed.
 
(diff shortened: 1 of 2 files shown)


2996_reflection/reflection.md
```diff
@@ -6767,6 +6767,100 @@ consteval source_location source_location_of(info r);
 
 ::: std
 ::: addu
+```cpp
+consteval bool $has-type$(info r); // exposition only
+```
+
+[#]{.pnum} *Returns*: `true` if  `r` represents a value, object, variable, function that is not a constructor or destructor, enumerator, non-static data member, bit-field, direct base class relationship, or data member description. Otherwise, `false`.
```
brevzin, line 6774: pre-existing: should be unnamed bit-field?

2996_reflection/reflection.md
```diff
@@ -6767,6 +6767,100 @@ consteval source_location source_location_of(info r);
...
+
+```cpp
+consteval info type_of(info r);
+```
+
+[#]{.pnum} *Constant When*: `$has-type$(r)` is `true`.
+
+[#]{.pnum} *Returns*:
+
+- [#.#]{.pnum} If `r` represents a value, object, variable, function, non-static data member, or bit-field, then the type of what is represented by `r`.
```
brevzin, line 6784: ditto unnamed bit-field

2996_reflection/reflection.md
katzdm: Done

2996_reflection/reflection.md
katzdm: Done

brevzin (approved):

PR #272, opened 2025-05-29
A few updates.

katzdm:


2996_reflection/reflection.md
```diff
@@ -3553,17 +3553,6 @@ In the following example, the set of potential results of the initializer of `n`
...
 :::
 
-Break bullet 4.1 into sub-bullets and modify it to cover splicing of functions:
-
-::: std
-- [4.1]{.pnum} A function is named by an expression or conversion [`$E$`]{.addu} if it is the selected member of an overload set ([basic.lookup], [over.match], [over.over]) in an overload resolution performed as part of forming that expression or conversion, unless it is a pure virtual function and [either the expression]{.rm}[\ ]{.addu}
-
-  - [#.#.#]{.pnum} [`$E$`]{.addu} is not an `$id-expression$` naming the function with an explicitly qualified name[,]{.addu}
-  - [[#.#.#]{.pnum} [`$E$`]{.addu} is a `$splice-expression$`,]{.addu} or
-  - [#.#.#]{.pnum} [the expression]{.rm} [`$E$`]{.addu} forms a pointer to member ([expr.unary.op]).
```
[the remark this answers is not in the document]
brevzin: Added https://github.com/brevzin/cpp_proposals/pull/208 I think.

2996_reflection/reflection.md
```diff
@@ -4930,6 +4919,21 @@ using alias = [:^^S::type:];    // OK, type-only context
...
+Prefer "designates", and disallow pointers to anonymous unions, in paragraph 2.
+
+::: std
+[2]{.pnum} In a declaration `T D` where `D` has the form
+
+```
+$nested-name-specifier$ * $attribute-specifier-seq$@~_opt_~@ $cv-qualifier-seq$@~_opt_~@ D1
+```
+
+and the `$nested-name-specifier$` [denotes]{.rm} [designates]{.addu} a class, and the type of the contained `$declarator-id$` in the declaration `T D1` is "_derived-declarator-type-list_ `T`", the type of the `$declarator-id$` in `D` is "_derived-declarator-type-list_ `$cv-qualifier-seq$` pointer to member of class `$nested-name-specifier$` of type `T`". The optional `$attribute-specifier-seq$` ([dcl.attr.grammar]) appertains to the pointer-to-member. [`T` shall not be an anonymous union.]{.addu}
```
brevzin, line 4933: Sorry why are we disallowing this? Cause like:
```cpp
struct C { union { int i; } };
```
Given a `c` it is possible to to to the union object with `c.[:the_union:]` and then to splice again to get to the `i` via `c.[:the_union:].[:the_i:]`. I actually can't tell if this is rejecting `&[:the_union:]` or `&[:the_i:]` (or rather, spelling of those types), but I'm not sure why to reject either?

katzdm: Pointers-to-members `MemTy Cls::*` can't be formed when `Cls` is an anonymous union, and there's some existing core issue stuff to clarify that. The only way to spell the type in the first place is with _splice-type-specifiers_, so I'm arguing that we should just make the type ill-formed in the first place.

katzdm: This is still fine:

```cpp
struct Cls {
  union {
    int mem;
  };
};

auto p = &[:^^Cls::mem:];  // type is 'int Cls::*'.
```

https://godbolt.org/z/EKjdhG8q7

brevzin: Okay so 
```cpp
constexpr auto r = parent_of(^^Cls::mem);
static_assert(is_union_type(r)); // yes
using A = int [:r:]::*; // bad
using B = [:r:]; // fine
using C = [:r:]*; // fine
```
?

brevzin: Aren't you rejecting the wrong type here? In `int [:r:]::*`, `T` is `int`.

brevzin (approved):

